PR #9, opened 2025-07-23
feat: `gastime` package

### LINKED ISSUE #11: `gastime` package production readiness

ARR4N:
Introduces the gas clock, an extension of a `proxytime.Time[gas.Gas]` that also tracks a "continuous" equivalent of ACP-176 gas excess at gas-unit resolution instead of per second.

Closes #11


gastime/gastime.go
```diff
@@ -0,0 +1,112 @@
...
+	return err
+}
+
+// Tick is equivalent to [proxytime.Time.Tick] except that it also updates the
+// gas excess.
+func (tm *Time) Tick(g gas.Gas) {
+	tm.Time.Tick(g)
+
+	R, T := tm.Rate(), tm.Target()
+	quo, _, _ := intmath.MulDiv(g, R-T, R) // overflow is impossible as (R-T)/R < 1
```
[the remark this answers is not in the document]
ARR4N: The alternative is to create an `intmath.MulDiv()` that doesn't perform the overflow check and just allows it to panic (which we know it won't). While I understand the AI's point, I'm not convinced that it's a true concern when the impossible error is the only thing that can be returned (unless there's a weird future refactoring of `MulDiv`).

gastime/marshal.go
```diff
@@ -0,0 +1,49 @@
...
+	"github.com/ava-labs/avalanchego/vms/components/gas"
+
+	"github.com/ava-labs/strevm/proxytime"
+)
+
+//go:generate go run github.com/StephenButtolph/canoto/canoto $GOFILE
+
+// A TimeMarshaler can marshal a time to and from canoto. It is of limited use
+// by itself and SHOULD only be used via a wrapping [Time].
+type TimeMarshaler struct { //nolint:tagliatelle // TODO(arr4n) submit linter bug report
```
StephenButtolph, line 17: For future reference, this is definitely a linter error:
```sh
$ golangci-lint run --config .golangci.yml ./...                       
gastime/marshal.go:17:20: unable to get field name: unexpected error: type *ast.IndexExpr: {"X":{"X":{"NamePos":118436797,"Name":"proxytime","Obj":null},"Sel":{"NamePos":118436807,"Name":"Time","Obj":null}},"Lbrack":118436811,"Index":{"X":{"NamePos":118436812,"Name":"gas","Obj":null},"Sel":{"NamePos":118436816,"Name":"Gas","Obj":null}},"Rbrack":118436819} (tagliatelle)
type TimeMarshaler struct { // TODO(arr4n) submit linter bug report
                   ^
```

gastime/gastime.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+func (tm *Time) establishInvariants() {
+	tm.Time.SetRateInvariants(&tm.target, &tm.excess)
+}
+
+// New returns a new [Time], set from a Unix timestamp. The consumption of
+// `2*target` units of [gas.Gas] is equivalent to a tick of 1 second.
+func New(unixSeconds uint64, target, startingExcess gas.Gas) *Time {
+	return makeTime(proxytime.New(unixSeconds, 2*target), target, startingExcess)
```
StephenButtolph, line 51: Here we multiply `target` by 2. In the current ACP-176 implementation, we clamp `R` to `MaxUint64` rather than overflowing. ref: https://github.com/ava-labs/coreth/blob/master/plugin/evm/upgrade/acp176/acp176.go#L101

gastime/gastime.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+// Excess returns the `x` variable of ACP-176.
+func (tm *Time) Excess() gas.Gas {
+	return tm.excess
+}
+
+// Price returns the price of a unit of gas, i.e. the "base fee".
+func (tm *Time) Price() gas.Price {
+	return gas.CalculatePrice(1 /* M */, tm.excess, 87*tm.target /* K */)
```
StephenButtolph, line 73: When calculating the conversion constant, we have an unchecked multiplication by 87 which is currently capped: https://github.com/ava-labs/coreth/blob/master/plugin/evm/upgrade/acp176/acp176.go#L93

gastime/gastime.go
```diff
@@ -0,0 +1,112 @@
...
+// FastForwardTo is equivalent to [proxytime.Time.FastForwardTo] except that it
+// may also update the gas excess.
+func (tm *Time) FastForwardTo(to uint64) {
+	sec, frac := tm.Time.FastForwardTo(to)
+	if sec == 0 && frac.Numerator == 0 {
+		return
+	}
+
+	R, T := tm.Rate(), tm.Target()
+	quo, _, _ := intmath.MulDiv(R*gas.Gas(sec)+frac.Numerator, T, R) // overflow is impossible as T/R < 1
```
StephenButtolph, line 110: I agree that `MulDiv` will never return an error, but it does seem possible that `R * sec + Numerator` could overflow.

While that probably isn't relevant for C-chain, it will likely be relevant for subnet-evm

StephenButtolph (commented):
Really my only questions/concerns are cases of potential overflow.

gastime/gastime.go
StephenButtolph: Talked offline, going to ensure R never overflows by limiting T to be `MaxUint64/2`

ARR4N: Done

gastime/gastime.go
```diff
@@ -0,0 +1,170 @@
...
+	target = clampTarget(target)
+	return makeTime(proxytime.New(unixSeconds, rateOf(target)), target, startingExcess)
+}
+
+// TargetToRate is the ratio between [Time.Target] and [proxytime.Time.Rate].
+const TargetToRate = 2
+
+// MaxTarget is the maximum allowable [Time.Target] to avoid overflows of the
+// associated [proxytime.Time.Rate]. Values above this are silently clamped.
+const MaxTarget = gas.Gas(math.MaxUint64 / TargetToRate)
```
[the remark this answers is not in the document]
ARR4N: Since it was realistically only ever going to be used as a `gas.Gas`, I did it to avoid having to set the type every time I wanted to use `require.Equal` (and `assert`) in testing because they're ~written for Python~ not type safe.

gastime/gastime.go
```diff
@@ -0,0 +1,170 @@
...
+	tm.excess += quo
+}
+
+// FastForwardTo is equivalent to [proxytime.Time.FastForwardTo] except that it
+// may also update the gas excess.
+func (tm *Time) FastForwardTo(to uint64) {
+	sec, frac := tm.Time.FastForwardTo(to)
+	if sec == 0 && frac.Numerator == 0 {
+		return
+	}
```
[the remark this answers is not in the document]
ARR4N: Yup, purely for optimisation as everything would otherwise be a no-op.

PR #18, opened 2025-09-12
feat: `blocks` package

### LINKED ISSUE #12: `blocks` package production readiness

ARR4N:
Introduces the `blocks.Block` type (and associated helpers), which satisfies the `adaptor.BlockProperties` interface introduced in #7. While the motivation of specific functionality may not be clear in isolation, context can be garnered from the prototype code (f79aff08) and how it uses this package in `saexec` and the primary SAE implementation.

Recommended review order:

1. `block.go` - general scaffolding and construction
2. `snow.go` - implementation of `adaptor.BlockProperties` methods
3. `execution.go` - transition from Accepted to Executed
4. `settlement.go` - transition from Executed to Settled; and also inspection of existing/hypothetical settlement (for block building)
5. ~`db.go` - database functionality used by the above, co-located for shared logic~
6. Other non-test files in `blocks` package

Tests can be reviewed either immediately after their respective non-test file, or batched at the end. Review the `cmputils` and `saetest` additions before reviewing any tests.

Closes #12.


blocks/snow.go
```diff
@@ -0,0 +1,53 @@
...
+// returns an error, it is logged at the WARNING level and a nil slice is
+// returned.
+func (b *Block) Bytes() []byte {
+	buf, err := rlp.EncodeToBytes(b)
+	if err != nil {
+		b.log.Warn("RLP encoding error", zap.Error(err))
+		return nil
+	}
+	return buf
+}
```
StephenButtolph, line 42: nit: Feels like the log should probably be an `Error`. Is this case ever expected?

blocks/settlement.go
```diff
@@ -0,0 +1,165 @@
...
+	a := b.ancestry.Load()
+	if a == nil {
+		b.log.Error(errBlockResettled.Error())
+		return fmt.Errorf("%w: block height %d", errBlockResettled, b.Height())
+	}
+	if !b.ancestry.CompareAndSwap(a, nil) { // almost certainly means concurrent calls to this method
+		b.log.Fatal("Block ancestry changed during settlement")
+		// We have to return something to keen the compiler happy, even though we
+		// expect the Fatal to be, well, fatal.
+		return errors.New("block ancestry changed during settlement")
```
StephenButtolph, line 34: nit: mind moving this out to a global? I know it should never be hit but seems nice to be consistent

blocks/settlement.go
```diff
@@ -0,0 +1,165 @@
...
+// of `x` and `y` (as described in [Block.Settles]). It is intended for use
+// during block building and defines `x` as the block height of
+// `b.LastSettled()` while `y` as the height of the argument passed to this
+// method.
+//
+// The argument is typically the return value of [LastToSettleAt], where that
+// function receives b as the parent. See the Example.
+func (b *Block) IfChildSettles(lastSettledOfChild *Block) []*Block {
+	return settling(b.LastSettled(), lastSettledOfChild)
+}
```
StephenButtolph, line 104: I was pretty confused by this function. After staring at it for long enough, I realized that it is returning what blocks the child will settle if it is accepted.

Can we rename this to `ChildSettles`?

Also can we comment this it is not valid to call `(If)ChildSettles` if `Block.MarkSettled` has been called on this block?

blocks/settlement.go
```diff
@@ -0,0 +1,165 @@
...
+			return block, true
+		}
+
+		// TODO(arr4n) more fine-grained checks are possible for scenarios where
+		// (a) `block` could never execute before `settleAt` so we would
+		// `continue`; and (b) `block` will definitely execute in time and
+		// `child` could never, in which case return `nil, false`.
+		_ = child
+
+		return nil, false
```
StephenButtolph, line 163: While this function does seem correct, I feel like we will probably need to address this TODO before putting this into production.

I think we should be able to perfectly implement this function without too much added complexity, right?

blocks/block.go
```diff
@@ -0,0 +1,124 @@
...
+	// usage. The pointer MAY be nil if execution is yet to commence.
+	executionExceededSecond atomic.Pointer[uint64]
+
+	executed chan struct{} // closed after `execution` is set
+	settled  chan struct{} // closed after `ancestry` is cleared
+
+	log logging.Logger
+}
+
+var inMemoryBlockCount atomic.Uint64
```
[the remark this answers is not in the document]
ARR4N: My original intent was just [observation during the full VM integration test](https://github.com/ava-labs/strevm/blob/f79aff08f39b58aa7cf4eeb1456ce4f979263020/integration_test.go#L241), which should (famous last words) be sufficient as a demonstration of there not being leaks as long as I convert it to a test and not just a log. WDYT? Are we going to regret not having it available when debugging an OOM in production 🤦 ?

blocks/execution.go
```diff
@@ -0,0 +1,158 @@
...
+// Executed reports whether [Block.MarkExecuted] has been called without
+// resulting in an error.
+func (b *Block) Executed() bool {
+	return b.execution.Load() != nil
+}
+
+func executionArtefact[T any](b *Block, desc string, get func(*executionResults) T) T {
+	e := b.execution.Load()
+	if e == nil {
+		b.log.Error(fmt.Sprintf("Get block %s before execution", desc))
```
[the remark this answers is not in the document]
ARR4N: I'm still not used to structured logging and when an attempt to call `b.log.Errorf(...` showed that it didn't exist I just assumed it was impossible. Bizarrely, the[ primary logging package](https://pkg.go.dev/github.com/golang/glog) we used at elGoog wasn't structured, and old habits die hard.

> As an aside, my spellcheck always hates reviewing your code, as it (and my own brain) are screaming to make this "artifact" lol

This was deliberate. I couldn't remember which was which so I did a search and then chose the non-American version as a means of peeing on the tree.

blocks/db.go
```diff
@@ -0,0 +1,125 @@
...
+	"github.com/ava-labs/libevm/core/types"
+	"github.com/ava-labs/libevm/ethdb"
+	"github.com/ava-labs/libevm/params"
+	"github.com/ava-labs/libevm/trie"
+)
+
+/* ===== Common =====*/
+
+func blockNumDBKey(prefix string, blockNum uint64) []byte {
+	return binary.BigEndian.AppendUint64([]byte(prefix), blockNum)
```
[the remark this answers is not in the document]
ARR4N: Done. It bugs me that you can't edit suggestions before committing them, so I pushed my own commit.

blocks/db.go
```diff
@@ -0,0 +1,125 @@
...
+}
+
+func (b *Block) writeToKVStore(w ethdb.KeyValueWriter, key func(uint64) []byte, val []byte) error {
+	return w.Put(key(b.NumberU64()), val)
+}
```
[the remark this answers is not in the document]
ARR4N: Both writes are used in the [post-shutdown database recovery](https://github.com/ava-labs/strevm/blob/f79aff08f39b58aa7cf4eeb1456ce4f979263020/db.go#L197), not of state sync necessarily. This feels like it may need a chat instead of comment-tennis, so let me know if you still have concerns and then we can just record the summary here.

blocks/block.go
```diff
@@ -0,0 +1,124 @@
...
+	// Overlord as a sign of our unwavering fealty. See [InMemoryBlockCount] for
+	// observability.
+	ancestry atomic.Pointer[ancestry]
+	// Non-nil i.f.f. [Block.MarkExecuted] or [Block.ResotrePostExecutionState]
+	// have returned without error.
+	execution atomic.Pointer[executionResults]
+
+	// See [Block.SetInterimExecutionTime for setting and [LastToSettleAt] for
+	// usage. The pointer MAY be nil if execution is yet to commence.
+	executionExceededSecond atomic.Pointer[uint64]
```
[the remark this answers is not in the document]
ARR4N: Done. I didn't want it to get out of sync if the usage changed, hence the linking, but I've added a high-level description in addition to the links.

blocks/snow.go
ARR4N: Done. Completely agree and have no idea why I used a lower level (this particular block of code is some of the earliest in the SAE implementation).

blocks/settlement.go
ARR4N: Done.

go.mod
```diff
@@ -1,17 +1,54 @@
...
+	github.com/shirou/gopsutil v3.21.11+incompatible // indirect
+	github.com/syndtr/goleveldb v1.0.1-0.20220614013038-64ee5596c38a // indirect
+	github.com/tklauser/go-sysconf v0.3.12 // indirect
+	github.com/tklauser/numcpus v0.6.1 // indirect
+	github.com/yusufpapurcu/wmi v1.2.2 // indirect
+	golang.org/x/sync v0.15.0 // indirect
+	rsc.io/tmplfunc v0.0.3 // indirect
+)
+
 require (
```
StephenButtolph, line 52: Why are there multiple indirect dependency blocks? I've seen go do this a couple times now and am quite confused 😕

ARR4N: 🤷

blocks/settlement.go
```diff
@@ -0,0 +1,165 @@
...
+
+// settling returns all the blocks after `lastOfParent` up to and including
+// `lastOfCurr`, each of which are expected to be the block last-settled by a
+// respective block-and-parent pair. It returns an empty slice if the two
+// arguments have the same block hash.
+func settling(lastOfParent, lastOfCurr *Block) []*Block {
+	var settling []*Block
+	// TODO(arr4n) abstract this to combine functionality with iterators
+	// introduced by @StephenButtolph.
+	for s := lastOfCurr; s.ParentBlock() != nil && s.Hash() != lastOfParent.Hash(); s = s.ParentBlock() {
```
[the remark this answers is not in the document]
ARR4N: I've added `saetest.NewTBLogger()` and use it by default on all blocks, resulting in a test error if there's a logged WARNING or higher. I'll now look into it a bit more because I agree with your logic and we should be able to remove the nil check.

blocks/settlement.go
ARR4N: Merged #19 to address this. Thank you.

> I think we should be able to perfectly implement this function without too much added complexity, right?

I'm not sure about _perfectly_ but getting much closer now. I added a new TODO to explain how to eke out the last bits of efficiency.

blocks/db.go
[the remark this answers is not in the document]
ARR4N: See https://github.com/ava-labs/strevm/pull/18#discussion_r2354763973 re deferring this.

PR #19, opened 2025-09-16
feat: `LastToSettleAt()` supports unknown grandchild execution time

StephenButtolph:
Was thinking about how best to convey what I expected to happen here and realized it would be easiest to just make a PR.


ARR4N (approved):
Nice! Thank you. I pushed a slight refactoring and updated the PR title to reflect what's happening.

PR #20, opened 2025-09-26
feat: `saexec` package

ARR4N:


blocks/blockstest/chain.go
```diff
@@ -28,12 +31,53 @@ func NewChainBuilder(genesis *blocks.Block) *ChainBuilder {
...
+	opts = slices.Concat(cb.opts, opts)
+
 	last := cb.Last()
-	eth := NewEthBlock(last.EthBlock(), txs, opts...)
-	cb.chain = append(cb.chain, NewBlock(tb, eth, last, nil)) // TODO(arr4n) support last-settled blocks
+	eth := NewEthBlock(last.EthBlock(), txs, ethBlockOptions(opts)...)
+	cb.chain = append(cb.chain, NewBlock(tb, eth, last, nil, blockOptions(opts)...)) // TODO(arr4n) support last-settled blocks
+
 	return cb.Last()
 }
```
StephenButtolph: optional (and pedantic), but do `ethBlockOptions` and `blockOptions` really improve readability?

I personally find the layers of indirection less clear:
```suggestion
// NewBlock constructs and returns a new block in the chain.
func (cb *ChainBuilder) NewBlock(tb testing.TB, txs []*types.Transaction, opts ...ChainOption) *blocks.Block {
tb.Helper()
	opts = slices.Concat(cb.opts, opts)
	options := options.ApplyTo(&chainOptions{}, opts...)

last := cb.Last()
	eth := NewEthBlock(last.EthBlock(), txs, options.ethOpts...)
	cb.chain = append(cb.chain, NewBlock(tb, eth, last, nil, options.saeOpts...)) // TODO(arr4n) support last-settled blocks

return cb.Last()
}
```

dummy/dummy.go
```diff
@@ -0,0 +1,35 @@
...
+	return engine{}
+}
+
+type (
+	chainContext struct{}
+	engine       struct{ consensus.Engine }
+)
+
+func (chainContext) Engine() consensus.Engine                    { return engine{} }
+func (chainContext) GetHeader(common.Hash, uint64) *types.Header { panic("unimplemented") }
```
StephenButtolph: Is this deadcode? It seems to me that this can currently be hit by the `BLOCKHASH` opcode.

hook/hook.go
```diff
@@ -0,0 +1,59 @@
...
+	"github.com/ava-labs/strevm/intmath"
+	saeparams "github.com/ava-labs/strevm/params"
+)
+
+// Points define user-injected hook points.
+type Points interface {
+	GasTarget(parent *types.Block) gas.Gas
+	SubSecondBlockTime(*types.Block) gas.Gas
+	BeforeBlock(params.Rules, *state.StateDB, *types.Block) error
+	AfterBlock(*state.StateDB, *types.Block, types.Receipts)
```
StephenButtolph: What is the rationale for allowing `BeforeBlock` to return an error but _not_ allowing `AfterBlock` to return an error?

saexec/execution.go
```diff
@@ -0,0 +1,193 @@
...
+	for {
+		select {
+		case <-e.quit:
+			return
+
+		case block := <-e.queue:
+			logger := e.log.With(
+				zap.Uint64("block_height", block.Height()),
+				zap.Uint64("block_time", block.BuildTime()),
+				zap.Any("block_hash", block.Hash()),
```
StephenButtolph: I try to avoid `zap.Any`.

```suggestion
				zap.Stringer("block_hash", block.Hash()),
```

There is almost always a better type to use.

saexec/execution.go
```diff
@@ -0,0 +1,193 @@
...
+		if err != nil {
+			// This almost certainly means that the worst-case block inclusion
+			// has a bug.
+			logger.Error(
+				"Transaction execution errored (not reverted)",
+				zap.Int("tx_index", ti),
+				zap.Stringer("tx_hash", tx.Hash()),
+				zap.Error(err),
+			)
+			continue
```
StephenButtolph: Should this `return err`? By continuing here, we are getting into a situation where we could be including transactions on chain that weren't able to process correctly.

Admittedly, if we are in this position we are in a really bad spot... But wouldn't we want to limit the amount of state corruption as much as possible?

saexec/execution.go
```diff
@@ -0,0 +1,193 @@
...
+		return common.Hash{}, fmt.Errorf("%T.Commit() at end of block %d: %w", scratch.statedb, blockNum, err)
+	}
+
+	db, err := state.New(root, e.stateCache, scratch.snaps)
+	if err != nil {
+		return common.Hash{}, err
+	}
+	scratch.statedb = db
+	return root, nil
+}
```
StephenButtolph: TODO: The handling of executionScratchSpace seems weird to me. I want to look into how this is structured.

saexec/subscription.go
```diff
@@ -0,0 +1,43 @@
...
+package saexec
+
+import (
+	"github.com/ava-labs/libevm/core"
+	"github.com/ava-labs/libevm/core/types"
+	"github.com/ava-labs/libevm/event"
+)
+
+func (e *Executor) sendPostExecutionEvents(b *types.Block, receipts types.Receipts) {
+	e.headEvents.Send(core.ChainHeadEvent{Block: b})
```
StephenButtolph: Just commenting that in the geth code they don't always public a chain head event (due to performance concerns). I'm not sure that is relevant for us... But I figured I'd point it out as a weird quirk that they do.

saexec/saexec.go
```diff
@@ -0,0 +1,139 @@
...
+func (e *Executor) LastExecuted() *blocks.Block {
+	return e.lastExecuted.Load()
+}
+
+// TimeNotThreadsafe returns a clone of the gas clock that times execution. It
+// is only safe to call when all blocks passed to [Executor.Enqueue]
+// have been executed, and is only intended for use in tests.
+func (e *Executor) TimeNotThreadsafe() *gastime.Time {
+	return e.gasClock.Clone()
+}
```
StephenButtolph: Is there a reason this is:
1. Exported
2. Defined in a production file

At least currently, it is only used is the test in this package.

hook/hook.go
```diff
@@ -0,0 +1,59 @@
...
+	pts.AfterBlock(sdb, b, rs)
+}
+
+// MinimumGasConsumption MUST be used as the implementation for the respective
+// method on [params.RulesHooks]. The concrete type implementing the hooks MUST
+// propagate incoming and return arguments unchanged.
+func MinimumGasConsumption(txLimit uint64) uint64 {
+	_ = (params.RulesHooks)(nil) // keep the import to allow [] doc links
+	return intmath.CeilDiv(txLimit, saeparams.Lambda)
+}
```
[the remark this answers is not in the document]
ARR4N: > Is this defined this way to allow the dependent repo to fully manage all the libevm registrations?

Yup. If we start doing `libevm` registrations here then we have to rewire the whole of `coreth` and `subnet-evm` as well as change SAE every time `libevm` changes.

> This feels a bit odd to me.

I agree to an extent, but not enough that I think it warrants the above changes. If you can think of an elegant alternative then I'm very happy to change.

dummy/dummy.go
ARR4N: Yup, you're right. Fixed and added extensive testing of contextual op codes.

saexec/execution.go
```diff
@@ -0,0 +1,192 @@
...
+			if err := e.execute(block, logger); err != nil {
+				logger.Error("Block execution failed", zap.Error(err))
+				return
+			}
+		}
+	}
+}
+
+type executionScratchSpace struct {
+	snaps   *snapshot.Tree
```
StephenButtolph: Why is `snaps` part of the `executionScratchSpace`? `snaps` seems more similar to `stateCache` than it is to `statedb`.

Specifically, we never actually reassign `snaps`, and it isn't tied to a single `execute` call.

saexec/saexec_test.go
```diff
@@ -0,0 +1,741 @@
...
+
+func TestImmediateShutdownNonBlocking(t *testing.T) {
+	newSUT(t, defaultHooks()) // calls [Executor.Close] in test cleanup
+}
+
+func TestExecutionSynchronisation(t *testing.T) {
+	ctx, sut := newSUT(t, defaultHooks())
+	e, chain := sut.Executor, sut.chain
+
+	for range uint64(10) {
```
StephenButtolph:
```suggestion
	for range 10 {
```

saexec/execution.go
```diff
@@ -0,0 +1,193 @@
...
+		case <-e.quit:
+			return errExecutorClosed
+		case <-ctx.Done():
+			return ctx.Err()
+		default:
+			// If this happens then increase the channel's buffer size.
+			e.log.Warn("Execution queue buffer too small")
+			runtime.Gosched()
+		}
+	}
```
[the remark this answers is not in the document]
ARR4N: Nor mine, but in the absence of an infinitely buffered channel I don't think we have a _simple_ fix that also honours context cancellation and closure of `e.quit`. We could always use `sink.Monitor` 😉

saexec/saexec_test.go
ARR4N: Ahh left over from before I had the `saetest.Wallet` to handle nonces for me. Thanks for catching this!

saexec/execution.go
ARR4N: From the early days of spiking on the PoC I was being pedantic about ownership and since only `stateCache` is documented as being thread safe I isolated the others. Now that we have much more reliable `blocks.Block.ParentBlock().PostExecutionStateRoot()`, the `StateDB` can be created as a local variable (as you suggest) so the scratch space is removed entirely.

saexec/execution.go
```diff
@@ -0,0 +1,192 @@
...
+				logger.Error("Block execution failed", zap.Error(err))
+				return
+			}
+		}
+	}
+}
+
+type executionScratchSpace struct {
+	snaps   *snapshot.Tree
+	statedb *state.StateDB
```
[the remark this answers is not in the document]
ARR4N: See [above](https://github.com/ava-labs/strevm/pull/20/files#r2539648288).

saexec/saexec.go
```diff
@@ -0,0 +1,135 @@
...
+	e.lastExecuted.Store(lastExecuted)
+	if err := e.init(); err != nil {
+		return nil, err
+	}
+
+	go e.processQueue()
+	return e, nil
+}
+
+func (e *Executor) init() error {
```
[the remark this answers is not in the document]
ARR4N: Removed along with the scratch space.

saexec/saexec.go
```diff
@@ -0,0 +1,135 @@
...
+	"github.com/ava-labs/strevm/hook"
+)
+
+// An Executor accepts and executes a [blocks.Block] FIFO queue.
+type Executor struct {
+	quit, done chan struct{}
+	log        logging.Logger
+	hooks      hook.Points
+
+	gasClock     *gastime.Time
```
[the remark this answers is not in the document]
ARR4N: Removed. Similarly to `StateDB`, I didn't originally have reliable `Block` methods so this is a hangover from before `ExecutedByGasTime()` existed.

saexec/saexec_test.go
```diff
@@ -0,0 +1,741 @@
...
+
+func TestMain(m *testing.M) {
+	goleak.VerifyTestMain(
+		m,
+		goleak.IgnoreCurrent(),
+		// Despite the call to [snapshot.Tree.Disable] in [Executor.Close], this
+		// still leaks at shutdown. This is acceptable as we only ever have one
+		// [Executor], which we expect to be running for the entire life of the
+		// process.
+		goleak.IgnoreTopFunction("github.com/ava-labs/libevm/core/state/snapshot.(*diskLayer).generate"),
```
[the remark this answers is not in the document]
ARR4N: There's a race in the upstream implementation (still there at tip, just moved to a different file). The `goleak` error includes:

```
[Goroutine 25 in state chan receive, with github.com/ava-labs/libevm/core/state/snapshot.(*diskLayer).generate on top of the stack:
github.com/ava-labs/libevm/core/state/snapshot.(*diskLayer).generate(0x140000d0360, 0x140000dd5c0)
	/.../go/pkg/mod/github.com/ava-labs/libevm@v1.13.15-0.20251112182915-1ec8741af98f/core/state/snapshot/generate.go:722 +0x51c
```

The blocking channel receive:

```go
dl.lock.Lock()
dl.genMarker = nil
close(dl.genPending)
dl.lock.Unlock()

// Someone will be looking for us, wait it out
abort = <-dl.genAbort // <------ L#722 referenced by goleak
abort <- nil
```

But `snapshot.Tree.Disable()` bails early (because `layer.genMarker` was marked as `nil` above) and never sends to `layer.genAbort`.

```go
layer.lock.RLock()
generating := layer.genMarker != nil // <------ Is happening after the above snippet unlocks
layer.lock.RUnlock()
if !generating {
	// Generator is already aborted or finished
	break
}
// If the base layer is generating, abort it
if layer.genAbort != nil {
	abort := make(chan *generatorStats)
	layer.genAbort <- abort
	<-abort
}
```

saexec/execution.go
```diff
@@ -0,0 +1,193 @@
...
+	// acceptance.
+	if last, curr := e.lastExecuted.Load().Height(), b.Height(); curr != last+1 {
+		return fmt.Errorf("executing blocks out of order: %d then %d", last, curr)
+	}
+
+	rules := e.chainConfig.Rules(b.Number(), true /*isMerge*/, b.BuildTime())
+
+	// Since `b` hasn't been executed, it definitely hasn't been settled, so we
+	// are guaranteed to have a non-nil parent available.
+	parent := b.ParentBlock()
```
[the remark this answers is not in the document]
ARR4N: Yup, absolutely a better check. Done.

PR #26, opened 2025-11-17
feat: `txgossip` package

ARR4N:
Introduces an integration point between `txpool.TxPool`, `saexec.Executor` and `gossip.BloomSet`, with a simple transaction-prioritisation mechanism. Any specific rules for when to reject a tx from the mempool will likely require some `libevm` mods so can follow later.

The consensus worst-case validation will still need to track "in-flight" txs as the `LegacyPool` will only remove them once included. An enqueueing subscription mechanism is added for this purpose.

### DIFF
--- a/txgossip/txgossip_test.go
+++ b/txgossip/txgossip_test.go
@@ -376,14 +376,18 @@ func TestAPIBackendSendTxSignatureMatch(_ *testing.T) {
 }
 
 func FuzzEffectiveGasTip(f *testing.F) {
-	rng := rand.New(rand.NewPCG(0, 0)) //nolint:gosec // CSPRNG is unnecessary for fuzzing
-	for range 100 {
-		var u [12]uint64
-		for i := range u {
-			u[i] = rng.Uint64()
+	// The goal of these seeds is to exercise all possible orderings of fee cap,
+	// tip cap and base fee, also including a nil base fee.
+	ints := []uint64{0, 1, 2}
+	for _, feeCap := range ints {
+		for _, tipCap := range ints {
+			for _, baseFee := range ints {
+				for _, nilBase := range []bool{true, false} {
+					const z = uint64(0)
+					f.Add(nilBase, feeCap, z, z, z, tipCap, z, z, z, baseFee, z, z, z)
+				}
+			}
 		}
-		nilBaseFee := rng.IntN(2) == 0
-		f.Add(nilBaseFee, u[0], u[1], u[2], u[3], u[4], u[5], u[6], u[7], u[8], u[9], u[10], u[11])
 	}
 
 	f.Fuzz(func(t *testing.T, nilB bool, f0, f1, f2, f3, t0, t1, t2, t3, b0, b1, b2, b3 uint64) {


txgossip/txgossip.go
```diff
@@ -0,0 +1,134 @@
...
+type Set struct {
+	Pool  *txpool.TxPool
+	bloom *gossip.BloomFilter
+}
+
+// NewSet returns a new [gossip.Set] and a cleanup function that MUST be called
+// to release resources.
+func NewSet(logger logging.Logger, pool *txpool.TxPool, bloom *gossip.BloomFilter, targetBloomElements int) (*Set, func() error) {
+	txs := make(chan core.NewTxsEvent)
+	sub := pool.SubscribeTransactions(txs, false)
```
StephenButtolph: It seems like this only receives txs upon mempool reorgs. I would have thought that the bloom filter should update as soon as we add the tx to the pool to avoid unnecessary gossip

txgossip/txgossip.go
```diff
@@ -0,0 +1,134 @@
...
+	done := make(chan struct{})
+	go maintainBloomFilter(logger, bloom, txs, sub.Err(), done, targetBloomElements)
+
+	return &Set{pool, bloom}, func() error {
+		sub.Unsubscribe()
+		<-done
+		return pool.Close()
```
StephenButtolph: I'm not sure that this warrants changing... But I was a bit surprised that this cleanup function closed the `pool`. I probably would have assumed this it only cleaned up logic specific to the `Set`.

I do feel like a comment on this function is needed at least to convey this information... If it were me, I would probably just expect the pool to be closed by the caller rather than implicitly here.

txgossip/txgossip.go
```diff
@@ -0,0 +1,134 @@
...
+// A Set is a [gossip.Set] wrapping a [txpool.TxPool]. Transactions MAY be added
+// to the pool directly, or via [Set.Add].
+type Set struct {
+	Pool  *txpool.TxPool
+	bloom *gossip.BloomFilter
+}
+
+// NewSet returns a new [gossip.Set] and a cleanup function that MUST be called
+// to release resources.
+func NewSet(logger logging.Logger, pool *txpool.TxPool, bloom *gossip.BloomFilter, targetBloomElements int) (*Set, func() error) {
```
StephenButtolph: When does it make sense for a cleanup function to be a method on the returned type vs an anonymous function?

I feel like it is more common for the cleanup to be done in a `Close` type of method, but I suppose contexts are a common counterpoint.

txgossip/blockchain.go
```diff
@@ -0,0 +1,108 @@
...
+	return bc.exec.LastEnqueued().Header()
+}
+
+func (bc *blockchain) GetBlock(hash common.Hash, number uint64) *types.Block {
+	return bc.blocks.EthBlock(hash, number)
+}
+
+func (bc *blockchain) StateAt(root common.Hash) (*state.StateDB, error) {
+	return state.New(root, bc.exec.StateCache(), nil)
+}
```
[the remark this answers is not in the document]
ARR4N: Intent is currently undefined. Sorry, I should have been clearer in what I meant in the PR description (i.e. I'm kicking that can down the road entirely):

> Any specific rules for when to reject a tx from the mempool will likely require some libevm mods so can follow later.

The p2p architecture and gossip was all so new to me that I found this PR relatively difficult, so wanted to focus on learning it and getting the wiring correct. I'll concentrate on things like validity when we have the `VM` to orchestrate all the moving parts.

txgossip/pushpull.go
```diff
@@ -0,0 +1,68 @@
...
+// Add is a wrapper around [txpool.TxPool.Add], exposed to accept transactions
+// over [gossip]. Populating the [Set] via this method does *not* result in new
+// transactions being push-gossiped, but they will be included in pull-gossip
+// responses.
+func (s *Set) Add(tx Transaction) error {
+	errs := s.Pool.Add([]*types.Transaction{tx.Transaction}, false, false)
+	for i, err := range errs {
+		if errors.Is(err, txpool.ErrAlreadyKnown) {
+			errs[i] = nil
+		}
```
StephenButtolph: Because we are replacing this error with `nil`, we are going to end up calling `s.bloom.Add` even if the tx was already known. This can result in us resetting our bloom filter more frequently than expected.

txgossip/priority.go
```diff
@@ -0,0 +1,54 @@
...
+type LazyTransaction struct {
+	*txpool.LazyTransaction
+	Sender common.Address
+}
+
+// TransactionsByPriority calls [txpool.TxPool.Pending] with the given filter,
+// collapses the results into a slice, and sorts said slice by decreasing gas
+// tip then chronologically. Transactions from the same sender are merely sorted
+// by increasing nonce.
+func (s *Set) TransactionsByPriority(filter txpool.PendingFilter) []*LazyTransaction {
```
StephenButtolph: I assume the plan is for this to be used by the block builder. fwiw the normal block builder includes an optimization to skip an entire account in the case that one of the txs is not valid. I'm not sure how important of an optimization this is, and it is also something we can add later as needed... So I don't think it's worth changing for now, but maybe we should add a TODO to investigate it?

txgossip/pushpull.go
```diff
@@ -0,0 +1,46 @@
...
+	// We filter out [txpool.ErrAlreadyKnown] in [txSet.Add] but want to keep it
+	// for RPCs.
+	if s.set.pool.Has(ethTx.Hash()) {
+		return txpool.ErrAlreadyKnown
+	}
+
+	tx := Transaction{ethTx}
+	if err := s.BloomSet.Add(tx); err != nil {
+		return err
+	}
```
StephenButtolph: Commenting on a behavior change here: normally we would mark these transactions as local on the API side. But this doesn't mark these as local txs.

I don't _think_ this is a problem... But figured it is worth a comment.

Typically (for an API node), I don't think we'd actually want to specify something as local... Since that permanently marks the account as local... But some people may want that behavior.

PR #30, opened 2025-11-19
test: `Executor.Close()` persists snapshot to disk

### LINKED ISSUE #29: Test persistence of `snapshot` to disk on `saexec.Executor.Close()`

ARR4N:
Closes #29, an open question after merging the `saexec.Executor`.

### DIFF
--- a/saexec/saexec_test.go
+++ b/saexec/saexec_test.go
@@ -759,8 +759,8 @@ func TestSnapshotPersistence(t *testing.T) {
 	// The crux of the test is whether we can recover the EOA nonce using only a
 	// new set of snapshots, recovered from the databases.
 	conf := snapshot.Config{
-		CacheSize:  128,
-		AsyncBuild: false,
+		CacheSize: 128,
+		NoBuild:   true, // i.e. MUST be loaded from disk
 	}
 	snaps, err := snapshot.New(conf, sut.db, e.StateCache().TrieDB(), last.PostExecutionStateRoot())
 	require.NoError(t, err, "snapshot.New(..., [post-execution state root of last-executed block])")


saexec/saexec.go
```diff
@@ -102,15 +102,12 @@ func (e *Executor) Close() {
 	// no-op, so we ignore it.
 	if root := e.LastExecuted().PostExecutionStateRoot(); root != e.snaps.DiskRoot() {
 		if err := e.snaps.Cap(root, 0); err != nil {
-			e.log.Warn(
-				"snapshot.Tree.Cap([last post-execution state root], 0)",
-				zap.Stringer("root", root),
-				zap.Error(err),
-			)
+			return fmt.Errorf("snapshot.Tree.Cap([last post-execution state root], 0): %v", err)
```
StephenButtolph:
```suggestion
			return fmt.Errorf("snapshot.Tree.Cap([last post-execution state root], 0): %w", err)
```

saexec/saexec_test.go
```diff
@@ -724,3 +730,48 @@ var _ = blockstest.ModifyHeader((*blockNumSaver)(nil).store)
...
+	// [newSUT] creates a cleanup that also calls [Executor.Close], which isn't
+	// valid usage. The simplest workaround is to just replace the quit channel
+	// so it can be closed again.
+	e.quit = make(chan struct{})
+
+	// The crux of the test is whether we can recover the EOA nonce using only a
+	// new set of snapshots, recovered from the databases.
+	conf := snapshot.Config{
+		CacheSize:  128,
+		AsyncBuild: false,
```
StephenButtolph: Even without the `Cap` call, this test passes. Specifically, it'll start generating a snapshot in the background. I think we should be specifying `NoBuild` so that if the snap isn't available then it errors.
```suggestion
		CacheSize: 128,
		NoBuild:   true,
```

StephenButtolph:
I'm not sure that this closes #29. (The comment I made on the executor PR was intended to be a bit more holistic than just the snapshot logic).

Specifically, (after talking with @alarso16). I think we need to be calling `triedb.Commit` to flush the trie nodes to disk during shutdown. Additionally, I think we will want to periodically be committing the triedb during execution to limit the amount of reexecution that is required upon a restart.

PR #33, opened 2025-11-28
fix(blocks): bugs and edge cases

ARR4N:
A collection of unrelated (sorry) but individually tiny fixes that I found while implementing block building:

1. In no longer embedding the geth `types.Block`, we broke RLP encoding. I'm not adding a test here because it will break every single integration test if not working properly so a unit test is redundant.

2. `saetest.EventCollector.WaitForAtLeast()` is now `Context`-aware so tests fail quickly when used in conjunction with `saetest.TBLogger.CancelOnError(context.Context)`.

3. Miscellaneous convenience and invariant assertions that are easier to understand by just seeing the code.

### DIFF
--- a/sae/http.go
+++ b/sae/http.go
@@ -20,8 +20,3 @@ func (vm *VM) CreateHandlers(context.Context) (map[string]http.Handler, error) {
 func (vm *VM) NewHTTPHandler(context.Context) (http.Handler, error) {
 	return nil, errUnimplemented
 }
-
-// CreateHTTP2Handler returns `(nil, nil)`.
-func (vm *VM) CreateHTTP2Handler(ctx context.Context) (http.Handler, error) {
-	return nil, nil
-}


saetest/saetest.go
```diff
@@ -85,10 +86,28 @@ func (c *EventCollector[T]) Unsubscribe() error {
...
 	c.cond.L.Lock()
 	defer c.cond.L.Unlock()
 	for len(c.all) < n {
 		c.cond.Wait()
+		if ctx.Err() != nil {
+			return context.Cause(ctx)
+		}
 	}
+	return nil
 }
```
StephenButtolph: This implementation of `WaitForAtLeast` does not actually guarantee context cancellation will result in waking up the goroutine.

(Shameless self plug): I'd recommend using the `lock.Cond` implementation here: https://github.com/ava-labs/avalanchego/blob/master/utils/lock/cond.go which has the same usage as the `sync.Cond` implementation but changes `Wait` to take in a context and return an error.

StephenButtolph: This implementation of `WaitForAtLeast` does not actually guarantee context cancellation will result in waking up the goroutine.

For example, add a `time.Sleep(time.Second)` immediately prior to calling `c.cond.L.Lock()`. The unit test will deadlock (aka timeout).

(Shameless self plug): I'd recommend using the `lock.Cond` implementation here: https://github.com/ava-labs/avalanchego/blob/master/utils/lock/cond.go which has the same usage as the `sync.Cond` implementation but changes `Wait` to take in a context and return an error.

PR #35, opened 2025-12-04
Add official EVM Block test data and execution

ceyonur:
This PR adds the test execution (from https://github.com/ava-labs/libevm/tree/main/tests) and `testdata` as a submodule from the repository https://github.com/ethereum/tests (at `fa51c5c164` as same commit in libevm@main). It only executes `BlockchainTests` and `LegacyTests` which are inserting test blockchains (in json files) with SAE's own blockbuilder/chain (through `SUT`) and then executes with `saexec`. Tests are passing but with few limitations and hacks (see `saexec/ethtests/block_test.go` and comments for that file) :

1- `GeneralStateTest` are too slow tried to run them but CI fails see [comment](https://github.com/ava-labs/strevm/pull/35#discussion_r2626364594)
2- We cannot run `InvalidBlocks` test because they're asserting either that block insertion is failing (which we don't have any verification yet), or they fail at execution (which we don't allow). We can revisit them once we add the verification.
3- There are some tests like `bcForkStressTest` and `bcMultiChainTest` testing multi-chains and forks with non-linear blocks and multiple possible chains. We currently don't support this in test chain builders. So these tests are skipped.
4- We don't have any execution pre-checks yet (like MaxCodeSize and intrinsic gas checks) so some tests are failing with a `FAIL` log in saexec. Once we have those checks we can execute those tests.
5- SAE has it's own baseFee applied to the header just before the execution, which means header's baseFee won't directly applied to the state where EVM tests expect it to be applied as it's; thus failing some state tests like balances, rewards etc. This is currently handled by faking the gas clock in `insertWithHeaderBaseFee`


blocks/blockstest/blocks.go
```diff
@@ -102,22 +107,18 @@ func WithLogger(l logging.Logger) BlockOption {
...
-		Config: config,
-		Alloc:  alloc,
-	}
-
 	tdb := state.NewDatabaseWithConfig(db, conf.tdbConfig).TrieDB()
-	_, hash, err := core.SetupGenesisBlock(db, tdb, gen)
+	_, _, err := core.SetupGenesisBlock(db, tdb, gen)
 	require.NoError(tb, err, "core.SetupGenesisBlock()")
-	require.NoErrorf(tb, tdb.Commit(hash, true), "%T.Commit(core.SetupGenesisBlock(...))", tdb)
+	// TODO(cey): This seems wrong, why are we committing the block hash rather than the root? We should already be committing that when we called SetupGenesisBlock.
```
ceyonur: We can probably remove these lines?

saexec/ethtests/block_test.go
```diff
@@ -0,0 +1,146 @@
...
+	// skip intrinsic gas tests
+	skippedTestRegexp = append(skippedTestRegexp, `.*/gas_usage.json/.*too_little_intrinsic_gas.*`)
+	bt.skipLoad(`.*/bcFrontierToHomestead/HomesteadOverrideFrontier.json`)
+	bt.skipLoad(`.*/bcFrontierToHomestead/ContractCreationFailsOnHomestead.json`)
+	// skip insufficient funds tests
+	skippedTestRegexp = append(skippedTestRegexp, `.*/use_value_in_tx.json/000-fork=Shanghai-tx_in_withdrawals_block`)
+	skippedTestRegexp = append(skippedTestRegexp, `.*/use_value_in_tx.json/002-fork=Cancun-tx_in_withdrawals_block`)
+	// Skip tx type check
+	bt.skipLoad(`.*/bcBerlinToLondon/initialVal.json`)
+	// TODO(cey): We cannot run baseFee related tests, since in SAE base fee calculation is different.
```
ceyonur: This was directly handled in the executor

saexec/ethtests/block_test.go
```diff
@@ -0,0 +1,146 @@
...
+				t.Skipf("test %s skipped", name)
+			}
+		}
+		execBlockTest(t, bt, test)
+	})
+}
+
+// TestExecutionSpecBlocktests runs the test fixtures from execution-spec-tests.
+func TestExecutionSpecBlocktests(t *testing.T) {
+	if !common.FileExist(executionSpecBlockchainTestDir) {
```
ceyonur: this is actually an empty file, so we don't run this test.

saexec/ethtests/block_test_util.go
```diff
@@ -0,0 +1,408 @@
...
+		// validate RLP decoding by checking all values against test file JSON
+		if err = validateHeader(b.BlockHeader, cb.Header()); err != nil {
+			return nil, fmt.Errorf("deserialised block header validation failed: %v", err)
+		}
+		validBlocks = append(validBlocks, b)
+		blocks = append(blocks, cb)
+	}
+
+	// Insert the blocks into the chain
+	i, err := sut.InsertChain(ctx, tb, blocks)
```
ceyonur: in the eth tests they were inserting a single block > execute > validate header and then continue to other blocks. Here we rather insert all collect validated blocks > enqueue > execute all of them and then return.

saexec/ethtests/block_test_util.go
```diff
@@ -0,0 +1,408 @@
...
+	Difficulty    *math.HexOrDecimal256
+	GasLimit      math.HexOrDecimal64
+	GasUsed       math.HexOrDecimal64
+	Timestamp     math.HexOrDecimal64
+	BaseFeePerGas *math.HexOrDecimal256
+	BlobGasUsed   *math.HexOrDecimal64
+	ExcessBlobGas *math.HexOrDecimal64
+}
+
+func (t *BlockTest) Run(tb testing.TB, snapshotter bool, scheme string, tracer vm.EVMLogger, postCheck func(error, *saexectest.SUT)) (result error) {
```
ceyonur: this is where the magic happens

saexec/saexectest/sut.go
ceyonur: I think this deserve to be it's own package and exported. I've also seen this in gossip tests, so we should probably use this there as well.

saexec/execution.go
```diff
@@ -112,7 +113,10 @@ func (e *Executor) execute(b *blocks.Block, logger logging.Logger) error {
 	}
 
 	header := types.CopyHeader(b.Header())
-	header.BaseFee = gasClock.BaseFee().ToBig()
+	// TODO(cey): this is a janky hack to pass the base fee tests in ethtests.
+	if !e.execOpts.preserveBaseFee {
+		header.BaseFee = gasClock.BaseFee().ToBig()
+	}
```
ceyonur: this is a very janky workaround. Without this I could not get any tests after `London` fork to pass. Because the baseFee changes both the end state, where tests expecting a certain balance after txs and also it changes rewards accumulated in coinbase.
Alternatives are:
1- Set the gas target before execution, so that we can try to make `gasClock` equivalent to whatever in the `header.BaseFee`
2- Modify tests and expected state.

blocks/blockstest/chain.go
```diff
@@ -84,6 +85,12 @@ func (cb *ChainBuilder) NewBlock(tb testing.TB, txs []*types.Transaction, opts .
 	return b
 }
 
+// Insert adds a block to the chain.
+func (cb *ChainBuilder) Insert(block *blocks.Block) {
```
ARR4N: `NewChainBuilder()` (for genesis) and `ChainBuilder.NewBlock()` should use `Insert()` so that all paths to block insertion are equivalent.

blocks/blockstest/chain.go
```diff
@@ -84,6 +85,12 @@ func (cb *ChainBuilder) NewBlock(tb testing.TB, txs []*types.Transaction, opts .
...
+	cb.chain = append(cb.chain, block)
+	cb.blocksByHash.Store(block.Hash(), block)
+}
```
ARR4N:
```suggestion
func (cb *ChainBuilder) Insert(tb testing.TB, b *blocks.Block) {
	tb.Helper()
	if len(cb.chain) > 0 {
		p := cb.Last()
		require.Equalf(tb, p.Height()+1, b.Height(), "%T.Insert() non-incrementing block height", cb)
		require.Equalf(tb, p.Hash(), b.ParentHash(), "%T.Insert() with mismatched parent hash", cb)
	}
	cb.chain = append(cb.chain, b)
	cb.blocksByHash.Store(b.Hash(), b)
}
```

Otherwise we risk breaking internal invariants.

blocks/blockstest/chain.go
```diff
@@ -114,3 +121,23 @@ func (cb *ChainBuilder) GetBlock(h common.Hash, num uint64) (*blocks.Block, bool
 	}
 	return b, true
 }
+
+// GetHashAtHeight returns the hash of the block at the given height, and a flag indicating if it was found.
+// If the height is greater than the number of blocks in the chain, it returns an empty hash and false.
+func (cb *ChainBuilder) GetHashAtHeight(num uint64) (common.Hash, bool) {
```
ARR4N: This is only ever used to pass the hash into `ChainBuilder.GetBlock()` so could just be `BlockByNumber(testing.TB, uint64) *blocks.Block` and similarly `GetNumberByHash()` could just be `BlockByHash(testing.TB, common.Hash) *blocks.Block`.

Since they don't need to conform to any particular signature (unlike `GetBlock()`, which is a `blocks.Source`) they can do the test failures internally if the block doesn't exist.

checksums.txt
```diff
@@ -0,0 +1,6 @@
+# This file contains sha256 checksums of optional build dependencies.
+
+# version:spec-tests 2.1.0
+# https://github.com/ethereum/execution-spec-tests/releases
+# https://github.com/ethereum/execution-spec-tests/releases/download/v2.1.0/
+ca89c76851b0900bfcc3cbb9a26cbece1f3d7c64a3bed38723e914713290df6c  fixtures_develop.tar.gz
```
ARR4N:
```suggestion
# sha256
# https://github.com/ethereum/execution-spec-tests/releases/download/v2.1.0/fixtures_develop.tar.gz
ca89c76851b0900bfcc3cbb9a26cbece1f3d7c64a3bed38723e914713290df6c  fixtures_develop.tar.gz

```

saexec/ethtests/block_test.go
ARR4N: Then why is it included in this PR?

saexec/ethtests/block_test_util.go
```diff
@@ -0,0 +1,441 @@
...
+		validBlocks = append(validBlocks, b)
+		blocks = append(blocks, cb)
+	}
+
+	// Insert the blocks into the chain
+	insertWithHeaderBaseFee(tb, sut, blocks)
+	return validBlocks, nil
+}
+
+func insertWithHeaderBaseFee(tb testing.TB, sut *SUT, bs types.Blocks) {
```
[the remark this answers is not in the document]
ARR4N: This is a useful primitive so please abstract `blockstest.WithFakeBaseFee(testing.TB, *blocks.Block, [other necessary args]) *blocks.Block`.

saexec/ethtests/block_test_util.go
```diff
@@ -0,0 +1,452 @@
...
+			AsyncBuild: false,
+		}
+		opts = append(opts, withSnapshotConfig(&snapshotConfig))
+	}
+	// Commit genesis state
+	gspec := t.genesis(config)
+	opts = append(opts, withGenesisSpec(gspec))
+
+	// Wrap the original engine within the beacon-engine
+	engine := beacon.New(ethash.NewFaker())
```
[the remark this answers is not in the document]
ceyonur: indeed, this is why we require #72. Otherwise we will always need to comply with Ethereum rules and specs/fixtures for testing these.

blocks/blockstest/chain.go
ceyonur: >However it seems this should be fine. I checked that in upstream by putting a panic here and seems there is no such case https://github.com/ava-labs/libevm/blob/main/core/blockchain.go#L1532-L1533.

On the second thought: This only checks if the given chain is ordered and linked, but it does not really check if it's a chain of single block (in a case we're inserting a single block). So tests actually expect a returned error (not failed test) in such [cases](https://github.com/ethereum/tests/blob/develop/BlockchainTests/InvalidBlocks/bcInvalidHeaderTest/wrongParentHash.json#L37).

As I said it won't fail in this PR, but it seems we need to return errors when we actually insert blocks (either in this `blockstest` package or through a prod-level chain builder)

saexec/ethtests/block_test_util.go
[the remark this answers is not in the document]
ceyonur: While #72 will be crucial for coreth/subnet-evm, SAE is also not 1-1 compliant with those JSON fixtures (baseFee, missing blob handling, beacon etc). We're currently patching those through upstream engines and before/after hooks with upstream-like handlings (difficulty, coinbase, beacon etc). 

I'm not sure what you mean by parsing JSON. We do parse them, create blocks in those JSONs and run blocks directly agains the executor ([here](https://github.com/ava-labs/strevm/blob/ceyonur/eth-tests/saexec/ethtests/block_test_util.go#L293-L294)). If you mean modifying JSON that will be generating fixtures (as described in #72).

blocks/blockstest/chain.go
ceyonur: I changed them but omitted the `TB` param, since we're calling these from the `readerAdapter` from `consensus.ChainHeaderReader `, which we cannot change the signature and take `TB`. We can set a `TB` for the `readerAdapter` when we create them but not sure if it's a good idea to carry it around.

PR #39, opened 2025-12-05
Update gas target immediately after block execution

StephenButtolph:
ACP176 expects the gas target to be updated immediately after the execution of the block. However, the prior code updates the target prior to executing the next block.

### DIFF
--- a/blocks/blockstest/blocks.go
+++ b/blocks/blockstest/blocks.go
@@ -104,7 +104,10 @@ func WithLogger(l logging.Logger) BlockOption {
 // marked as both executed and synchronous.
 func NewGenesis(tb testing.TB, db ethdb.Database, config *params.ChainConfig, alloc types.GenesisAlloc, opts ...GenesisOption) *blocks.Block {
 	tb.Helper()
-	conf := options.ApplyTo(&genesisConfig{}, opts...)
+	conf := &genesisConfig{
+		target: math.MaxUint64,
+	}
+	options.ApplyTo(conf, opts...)
 
 	gen := &core.Genesis{
 		Config: config,
@@ -124,14 +127,11 @@ func NewGenesis(tb testing.TB, db ethdb.Database, config *params.ChainConfig, al
 
 type genesisConfig struct {
 	tdbConfig *triedb.Config
-	target    *gas.Gas
+	target    gas.Gas
 }
 
 func (gc *genesisConfig) gasTarget() gas.Gas {
-	if gc.target == nil {
-		return math.MaxUint64
-	}
-	return *gc.target
+	return gc.target
 }
 
 // A GenesisOption configures [NewGenesis].
@@ -147,6 +147,6 @@ func WithTrieDBConfig(tc *triedb.Config) GenesisOption {
 // WithGasTarget overrides the gas target used by [NewGenesis].
 func WithGasTarget(target gas.Gas) GenesisOption {
 	return options.Func[genesisConfig](func(gc *genesisConfig) {
-		gc.target = &target
+		gc.target = target
 	})
 }


blocks/blockstest/blocks.go
```diff
@@ -114,13 +116,21 @@ func NewGenesis(tb testing.TB, db ethdb.Database, config *params.ChainConfig, al
...
 }
 
 type genesisConfig struct {
 	tdbConfig *triedb.Config
+	target    *gas.Gas
+}
+
+func (gc *genesisConfig) gasTarget() gas.Gas {
+	if gc.target == nil {
+		return math.MaxUint64
```
StephenButtolph: The default is changed from `1` to `Max` in this PR. Previously the value here wasn't really used because the gas target was fetched via the hook for the genesis, which was overridden later. But now we expect the target to be set correctly.

hook/hook_test.go
```diff
@@ -0,0 +1,67 @@
...
+	"testing"
+
+	"github.com/ava-labs/avalanchego/vms/components/gas"
+	"github.com/ava-labs/libevm/core/types"
+	"github.com/ava-labs/libevm/params"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/ava-labs/strevm/gastime"
+	. "github.com/ava-labs/strevm/hook"
```
ARR4N: (nit) I'll reconfigure the linter later, but please put this in its own group, at the end. It fundamentally changes how the file should be interpreted, so I think it should be made more obvious to a reader.

saexec/saexec_test.go
```diff
@@ -374,74 +374,83 @@ func TestGasAccounting(t *testing.T) {
...
 		{
-			target:          5 * gasPerTx,
+			// Initially set the gasTarget for the next block.
+			blockTime:       0,
+			numTxs:          0,
+			targetAfter:     5 * gasPerTx,
+			wantExecutedBy:  at(0, 0, 10*gasPerTx),
+			wantExcessAfter: 0,
+			wantPriceAfter:  1,
+		},
```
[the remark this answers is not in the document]
ARR4N: (thinking out loud) Practically, for `TransitionVM`, I think this means that the SAE VM will need to accept a last-synchronous block (it already does) and then, when initialising its gas clock, will set the target based on calling the `GasTargetAfter(lastSynchronous)`.

saexec/saexec_test.go
```diff
@@ -374,74 +374,83 @@ func TestGasAccounting(t *testing.T) {
...
 			wantExcessAfter: (12 + 20) * gasPerTx / 2,
 			wantPriceAfter:  1,
 		},
 		{
-			target:          5 * gasPerTx,
 			blockTime:       7, // fast-forward equivalent of 8 txs
 			numTxs:          16,
-			wantExecutedBy:  at(8, 6, 10*gasPerTx),
-			wantExcessAfter: (12 + 20 - 8 + 16) * gasPerTx / 2,
+			targetAfter:     10 * gasPerTx,           // double gas/block --> halve ticking rate
```
ARR4N: (note to self and other reviewers) Confirmed that this is moved here from the next step, which makes sense for a change from `target` to `targetAfter`.

saexec/saexec_test.go
```diff
@@ -374,74 +374,83 @@ func TestGasAccounting(t *testing.T) {
...
+			wantExecutedBy:  at(8, 6*2, 20*gasPerTx), // ending point scales
+			wantExcessAfter: 2 * (12 + 20 - 8 + 16) * gasPerTx / 2,
```
ARR4N:
```suggestion
			// Doubling the target scales both the ending time and excess to compensate.
			wantExecutedBy:  at(8, 2*6, 2*10*gasPerTx),
			wantExcessAfter: 2 * (12 + 20 - 8 + 16) * gasPerTx / 2,
```

I know that the `// starting point scales` comment was originally mine and it inspired the equivalent one here, but the fact that I as the original author really had to think meant that it was inadequate 🤦

PR #40, opened 2025-12-05
Remove hook.BeforeBlock and hook.AfterBlock

StephenButtolph:
In implementing the worstcase gas pricing, I want to share the gastime updating logic as much as possible. However, that should not result in calling `hook.Points.BeforeExecutingBlock` or `hook.Points.AfterExecutingBlock`.

This PR separates the gastime changes and also includes fuzzing to ensure that updating the target can not invalidate the worst case gas price assumptions.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -48,7 +48,7 @@ linters-settings:
     sections:
       - standard
       - default
-      - localmodule
+      - prefix(github.com/ava-labs/strevm)
       # The rest of these break developer expections, in increasing order of
       # divergence, so are at the end to increase the chance of being seen.
       - dot
--- a/blocks/blockstest/blocks.go
+++ b/blocks/blockstest/blocks.go
@@ -104,7 +104,10 @@ func WithLogger(l logging.Logger) BlockOption {
 // marked as both executed and synchronous.
 func NewGenesis(tb testing.TB, db ethdb.Database, config *params.ChainConfig, alloc types.GenesisAlloc, opts ...GenesisOption) *blocks.Block {
 	tb.Helper()
-	conf := options.ApplyTo(&genesisConfig{}, opts...)
+	conf := &genesisConfig{
+		target: math.MaxUint64,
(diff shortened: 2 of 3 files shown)


gastime/acp176.go
ARR4N: This precludes the `hook` package ever importing the `blocks` package. Not a problem for now, but flagging it so you're thinking about it. This will be OK if hooks only ever deal with Eth types, but that might not always be the case.

gastime/acp176_test.go
```diff
@@ -0,0 +1,133 @@
...
+	assert.Equal(t, expectedEndTime, tm.Unix(), "Unix time advanced by AfterBlock()")
+	assert.Equal(t, newTarget, tm.Target(), "Target updated by AfterBlock()")
+	assert.GreaterOrEqual(t, tm.Price(), enforcedPrice, "Price should not decrease in AfterBlock()")
+}
+
+func FuzzWorstCasePrice(f *testing.F) {
+	f.Fuzz(func(
+		t *testing.T,
+		initTimestamp, initTarget, initExcess,
+		time0, timeFrac0, used0, limit0, target0,
```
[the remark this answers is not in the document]
StephenButtolph: not soon enough

gastime/acp176_test.go
```diff
@@ -0,0 +1,138 @@
...
+	"github.com/ava-labs/libevm/core/types"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/ava-labs/strevm/hook/hookstest"
+)
+
+// TestTargetUpdateTiming verifies that the gas target is modified in AfterBlock
+// rather than BeforeBlock.
+func TestTargetUpdateTiming(t *testing.T) {
```
ARR4N: > [!NOTE]
> Not re-reviewing this and trusting you to update it if and when you change the base branch where it's introduced.

PR #41, opened 2025-12-05
Add `worstcase.State`

StephenButtolph:
This PR introduces an initial pass of worst-case state metering.

Worstcase state is intended to be used during block building (and verification) to ensure that the block can be included based on the queue, as well as ensuring that all included transactions will be able to be executed according to the EVM rules.

### DIFF
--- a/worstcase/state.go
+++ b/worstcase/state.go
@@ -196,8 +196,10 @@ func (s *State) ApplyTx(tx *types.Transaction) error {
 	}
 
 	// While EOA enforcement is not possible to guarantee in worst-case
-	// execution, we can prevent most cases here. Therefore, we must still
-	// handle non-EOA issuance later during actual execution.
+	// execution, we can prevent most cases here.
+	//
+	// TODO: We must still handle non-EOA issuance later during actual
+	// execution.
 	if codeHash := s.db.GetCodeHash(from); codeHash != (common.Hash{}) && codeHash != types.EmptyCodeHash {
 		return fmt.Errorf("%w: address %v, codehash: %s", core.ErrSenderNoEOA, from.Hex(), codeHash)
 	}
--- a/worstcase/state_test.go
+++ b/worstcase/state_test.go
@@ -220,7 +220,7 @@ func TestTransactionValidation(t *testing.T) {
 	require.NoError(t, err, "crypto.HexToECDSA() for eip3607Key")
 	eip3607EOA := crypto.PubkeyToAddress(eip3607Key.PublicKey)
 	eip3607Alloc := types.Account{
-		Balance: big.NewInt(1000000000000000000), // 1 ether
+		Balance: big.NewInt(params.Ether),
 		Nonce:   0,
 		Code:    common.Hex2Bytes("B0B0FACE"),
 	}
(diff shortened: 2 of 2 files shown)


worstcase/state.go
```diff
@@ -0,0 +1,250 @@
...
+
+var (
+	errNonConsecutiveBlocks = errors.New("non-consecutive block numbers")
+	errQueueFull            = errors.New("queue full")
+)
+
+// StartBlock updates the worst-case state to the beginning of the provided
+// block.
+//
+// It is not necessary for [types.Header.GasLimit] or [types.Header.BaseFee] to
```
ARR4N:
```suggestion
// It is not necessary for [types.Header.GasLimit] nor [types.Header.BaseFee] to
```

worstcase/state.go
```diff
@@ -0,0 +1,250 @@
...
+// be set.
+//
+// If the queue is too full to accept another block, an error is returned.
+func (s *State) StartBlock(hdr *types.Header) error {
+	if c := s.curr; c != nil {
```
ARR4N: This check will be unnecessary (and hence the internal one will be stronger) if we accept the hypothetically last-settled block in `NewState()`.

worstcase/state.go
```diff
@@ -0,0 +1,266 @@
...
+	errNonConsecutiveBlocks = errors.New("non-consecutive blocks")
+	errQueueFull            = errors.New("queue full")
+)
+
+// StartBlock updates the worst-case state to the beginning of the provided
+// block.
+//
+// It is not necessary for [types.Header.GasLimit] nor [types.Header.BaseFee] to
+// be set. However, all other fields should be populated and
+// [types.Header.ParentHash] must match the previous block's hash.
```
StephenButtolph: > all other fields should be populated

I say this because we end up passing this header into `hooks.SubSecondBlockTime`.

worstcase/state.go
```diff
@@ -0,0 +1,266 @@
...
+func (s *State) BaseFee() *uint256.Int {
+	return s.baseFee
+}
+
+type (
+	// AccountDebit includes an amount that an account should have debited,
+	// along with the nonce used to debit the account.
+	AccountDebit = hook.AccountDebit
+	// Op is an operation that can be applied to a [State].
+	Op = hook.Op
```
ceyonur: I'm comparing this with https://github.com/ava-labs/libevm/blob/main/core/state_transition.go#L274, most of them are checked in `ValidateTransaction`. The ones I could not see are:
1- Do we need to check if the sender is eoa? (https://github.com/ava-labs/libevm/blob/main/core/state_transition.go#L293)
2- Should we check the msg.Value? https://github.com/ava-labs/libevm/blob/main/core/state_transition.go#L407-L414

worstcase/state_test.go
```diff
@@ -0,0 +1,459 @@
...
+
+	err := state.StartBlock(&types.Header{
+		ParentHash: genesisHash,
+	})
+	require.NoError(t, err, "StartBlock()")
+
+	err = state.StartBlock(&types.Header{
+		ParentHash: genesisHash, // Should be the previously provided header's hash
+	})
+	require.ErrorIs(t, err, errNonConsecutiveBlocks, "nonconsecutive StartBlock()")
```
ARR4N:
```suggestion
	require.ErrorIs(t, err, errNonConsecutiveBlocks, "non-consecutive StartBlock()")
```

worstcase/state.go
StephenButtolph: I ended up adding the EOA check.

PR #42, opened 2025-12-09
ci: fuzz all the things

ARR4N:
Adds a required CI job `go_fuzz`, which runs on a matrix strategy fed by another job that finds all Go fuzz targets. When on `main` (e.g. after merging a PR) the `-fuzztime` is 15 minutes, otherwise (e.g. an open PR) it is 60 seconds. This pattern results in parallel fuzzing of all targets.


ARR4N:
> For future I recommend ensuring CI jobs invoke tasks (ala avalanchego) to ensure easy local reproducibility, but this is fine for now.

Ahh, that's why you do it like that!

PR #47, opened 2025-12-15
Add EndOfBlockOps hook

StephenButtolph:
AtomicTxs are separate transactions (completely unrelated to the ethereum transactions) which are stored in a libevm extension to the block body.

The AtomicTxs have the ability to spend native tokens from an EOA and to mint new tokens from cross-chain transfers. Additionally, they include multi-coin transfers, which will be implemented by the `AfterExecutingBlock` callback.

Each `Op` is going to correspond to a single atomic tx, which is why a slice is returned.

### DIFF
--- a/hook/hook.go
+++ b/hook/hook.go
@@ -34,10 +34,11 @@ type Points interface {
 	// For example, if the block timestamp is 10.75 seconds and the gas rate is
 	// 100 gas/second, then this method should return 75 gas.
 	SubSecondBlockTime(gasRate gas.Gas, h *types.Header) gas.Gas
-	// ExtraBlockOps returns operations outside of the normal EVM state changes
-	// to perform while executing the block. These operations will be performed
-	// during both worst-case and actual execution.
-	ExtraBlockOps(*types.Block) []Op
+	// EndOfBlockOps returns operations outside of the normal EVM state changes
+	// to perform while executing the block, after regular EVM transactions.
+	// These operations will be performed during both worst-case and actual
+	// execution.
+	EndOfBlockOps(*types.Block) []Op
 	// BeforeExecutingBlock is called immediately prior to executing the block.
 	BeforeExecutingBlock(params.Rules, *state.StateDB, *types.Block) error
 	// AfterExecutingBlock is called immediately after executing the block.
--- a/saexec/execution.go
+++ b/saexec/execution.go
@@ -163,7 +163,7 @@ func (e *Executor) execute(b *blocks.Block, logger logging.Logger) error {
 		receipts[ti] = receipt
 	}
 
-	for i, o := range e.hooks.ExtraBlockOps(b.EthBlock()) {
+	for i, o := range e.hooks.EndOfBlockOps(b.EthBlock()) {
 		blockGasConsumed += o.Gas
 		perTxClock.Tick(o.Gas)
 		b.SetInterimExecutionTime(perTxClock)
--- a/saexec/saexec_test.go
+++ b/saexec/saexec_test.go
@@ -345,7 +345,7 @@ func TestExecution(t *testing.T) {
 	})
 }
 
-func TestExtraBlockOps(t *testing.T) {
+func TestEndOfBlockOps(t *testing.T) {
 	hooks := defaultHooks()
 	ctx, sut := newSUT(t, hooks)
 	wallet := sut.wallet
--- a/hook/hookstest/stub.go
+++ b/hook/hookstest/stub.go
@@ -33,8 +33,8 @@ func (s *Stub) SubSecondBlockTime(gas.Gas, *types.Header) gas.Gas {
 	return s.SubSecondTime
 }
 
-// ExtraBlockOps ignores its argument and always returns [Stub.Ops].
-func (s *Stub) ExtraBlockOps(*types.Block) []hook.Op {
+// EndOfBlockOps ignores its argument and always returns [Stub.Ops].
+func (s *Stub) EndOfBlockOps(*types.Block) []hook.Op {
 	return s.Ops
 }
 


hook/hook.go
```diff
@@ -8,15 +8,64 @@
...
+	// GasPrice this operation is willing to spend
+	GasPrice uint256.Int
+	// From specifies the set of accounts and the authorization of funds to be
+	// removed from the accounts.
+	From map[common.Address]AccountDebit
+	// To specifies the amount to increase account balances by. These funds are
+	// not necessarily tied to the funds consumed in the From field. The sum of
+	// the To amounts may even exceed the sum of the From amounts.
+	To map[common.Address]uint256.Int
+}
```
StephenButtolph: If now isn't the time/place, we can change the `uint256.Int` into `*uint256.Int`... But it is probably faster and less error prone to pass these as values (as then they act exactly like `uint64`).

hook/hook.go
```diff
@@ -8,15 +8,72 @@
...
+
+var errInsufficientFunds = errors.New("insufficient funds")
+
+// ApplyTo applies the operation to the statedb.
+func (o *Op) ApplyTo(stateDB *state.StateDB) error {
+	for from, acc := range o.From {
+		// We use the state as the source of truth for the current nonce rather
+		// than the value provided by the hook. This prevents any situations,
```
ARR4N: Why is there a nonce in the hook if it doesn't do anything?

hook/hook.go
```diff
@@ -8,15 +8,73 @@
...
+// AccountDebit includes an amount that an account should have debited,
+// along with the nonce used to debit the account.
+type AccountDebit struct {
+	Nonce  uint64
+	Amount uint256.Int
+}
+
+// Op is an operation that can be applied to state during the execution of a
+// block.
+type Op struct {
```
ARR4N: I think an `Op` have an `ids.ID` for logging and other debugging purposes. WDYT?

hook/hook.go
```diff
@@ -8,15 +8,73 @@
...
+	To map[common.Address]uint256.Int
+}
+
+// ApplyTo applies the operation to the statedb.
+//
+// If an account has insufficient funds, [core.ErrInsufficientFunds] is returned
+// and the statedb is unchanged.
+func (o *Op) ApplyTo(stateDB *state.StateDB) error {
+	for from, acc := range o.From {
+		if balance := stateDB.GetBalance(from); balance.Lt(&acc.Amount) {
```
```suggestion
		if b := stateDB.GetBalance(from); b.Lt(&acc.Amount) {
```

(nit, style) The full word only adds noise without any signal as it's made redundant by the name of the method on `StateDB`. [Background](https://google.github.io/styleguide/go/decisions#:~:text=The%20general%20rule%20of%20thumb%20is%20that%20the%20length%20of%20a%20name%20should%20be%20proportional%20to%20the%20size%20of%20its%20scope%20and%20inversely%20proportional%20to%20the%20number%20of%20times%20that%20it%20is%20used%20within%20that%20scope.).

hook/hook.go
```diff
@@ -8,15 +8,73 @@
...
+	}
+	for from, acc := range o.From {
+		// We use the state as the source of truth for the current nonce rather
+		// than the value provided by the hook. This prevents any situations,
+		// such as with delegated accounts, where nonces might not be
+		// incremented properly.
+		//
+		// If overflow would have occurred here, the nonce must have already
+		// been increased during execution, so we are already protected against
+		// replay attacks.
```
ARR4N: > the nonce must have already been increased during execution

Please explain. To me, this implies that every `Op` is tied to a regular `types.Transaction`.

hook/hook_test.go
```diff
@@ -0,0 +1,136 @@
...
+						Amount: *uint256.NewInt(100_000),
+					},
+				},
+			},
+			wantErr: core.ErrInsufficientFunds,
+		},
+	}
+
+	db, err := state.New(types.EmptyRootHash, state.NewDatabase(rawdb.NewMemoryDatabase()), nil)
+	require.NoError(t, err, "state.New([empty root], [fresh memory db])")
```
ARR4N:
```suggestion
	_, _, db := ethtest.NewEmptyStateDB(t)
```

You'll need to import [github.com/ava-labs/libevm/libevm/ethtest](https://github.com/ava-labs/libevm/blob/9021836d7dbc55a91f74e5aca927b15ccce5f0cd/libevm/ethtest/evm.go#L38)

saexec/execution.go
```diff
@@ -162,6 +162,23 @@ func (e *Executor) execute(b *blocks.Block, logger logging.Logger) error {
...
+
+	for i, o := range e.hooks.ExtraBlockOps(b.EthBlock()) {
+		blockGasConsumed += o.Gas
+		perTxClock.Tick(o.Gas)
+		b.SetInterimExecutionTime(perTxClock)
+
+		if err := o.ApplyTo(stateDB); err != nil {
+			logger.Fatal(
+				"Extra block operation errored; see emergency playbook",
+				zap.Int("op_index", i),
```
ARR4N: An `Op.ID` would be useful here.

hook/hook.go
ceyonur: question: which execution we're referring here? Why is it impossible to have the overflow?

ceyonur: Should we still regardless check this in the previous loop and return err if there could be a overflow?

hook/hook_test.go
```diff
@@ -0,0 +1,136 @@
...
+	db.SetNonce(eoaMaxNonce, math.MaxUint64)
+	for _, tt := range tests {
+		require.ErrorIs(t, tt.op.ApplyTo(db), tt.wantErr, "ApplyTo %s", tt.name)
+		for _, acct := range tt.wantAccounts {
+			assert.Equalf(t, acct.nonce, db.GetNonce(acct.address), "nonce of account %s after %s", acct.address, tt.name)
+			assert.Equalf(t, acct.balance, db.GetBalance(acct.address), "balance of account %s after %s", acct.address, tt.name)
+		}
+		if t.Failed() {
```
ceyonur: what does this do? I think this makes sense if we run testcases in parallel, but I don't know what this currently does.

hook/hook_test.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	db, err := state.New(types.EmptyRootHash, state.NewDatabase(rawdb.NewMemoryDatabase()), nil)
+	require.NoError(t, err, "state.New([empty root], [fresh memory db])")
+	db.SetNonce(eoaMaxNonce, math.MaxUint64)
+	for _, tt := range tests {
+		require.ErrorIs(t, tt.op.ApplyTo(db), tt.wantErr, "ApplyTo %s", tt.name)
+		for _, acct := range tt.wantAccounts {
+			assert.Equalf(t, acct.nonce, db.GetNonce(acct.address), "nonce of account %s after %s", acct.address, tt.name)
+			assert.Equalf(t, acct.balance, db.GetBalance(acct.address), "balance of account %s after %s", acct.address, tt.name)
```
ceyonur: should these be `require` over `assert`s or are we running thise in parallel (that I don't see)

hook/hook.go
```diff
@@ -28,12 +34,56 @@ type Points interface {
...
+func (o *Op) ApplyTo(stateDB *state.StateDB) error {
+	for from, amount := range o.Burn {
+		if b := stateDB.GetBalance(from); b.Lt(&amount) {
+			return core.ErrInsufficientFunds
+		}
+	}
+	for from, amount := range o.Burn {
+		// If overflow would have occurred here, the nonce must have already
+		// been increased by a delegated account's execution, so we are already
+		// protected against replay attacks.
```
ARR4N: (note to self) Reason through this part again.

saexec/saexec_test.go
```diff
@@ -80,7 +80,7 @@ func newSUT(tb testing.TB, hooks hook.Points) (context.Context, SUT) {
 
 	wallet := saetest.NewUNSAFEWallet(tb, 1, types.LatestSigner(config))
 	alloc := saetest.MaxAllocFor(wallet.Addresses()...)
-	genesis := blockstest.NewGenesis(tb, db, config, alloc, blockstest.WithTrieDBConfig(tdbConfig))
+	genesis := blockstest.NewGenesis(tb, db, config, alloc, blockstest.WithTrieDBConfig(tdbConfig), blockstest.WithGasTarget(hooks.Target))
```
ARR4N: Ahh ok, I see why `hook.Points` wouldn't work here. While it's not ideal to force using the concrete type, generalising it at this point would be YAGNI so I'm happy with this change.

ARR4N (approved):

PR #48, opened 2025-12-16
feat: block build-verify-accept loop (without settlement)

ARR4N:
Implements `VM.{Build,Verify}Block()` based on shared logic but placeholder tx verification to be replaced by the `worstcase` package. Plumbs the necessary building blocks (e.g. mempool) to exercise these two primary methods along with a rudimentary `VM.AcceptBlock()` that does nothing more than push the block to the `Executor`.

Miscellaneous but self-contained improvements are littered amongst the primary changes because any TODO would have been just as intrusive.

### DIFF
--- a/sae/always.go
+++ b/sae/always.go
@@ -33,5 +33,7 @@ func (vm *SinceGenesis) Initialize(
 	fxs []*snowcommon.Fx,
 	appSender snowcommon.AppSender,
 ) error {
+	// TODO(arr4n) when implementing this, also consolidate [NewVM] and
+	// [VM.Init] so a freshly constructed [VM] is ready to use.
 	return errUnimplemented
 }


blocks/settlement.go
```diff
@@ -205,12 +207,16 @@ func LastToSettleAt(settleAt uint64, parent *Block) (b *Block, ok bool) {
 	// settled (see invariant in [Block]). If a block was already settled then
 	// only it or a later (i.e. unsettled) block can be returned by this loop,
 	// therefore we have a guarantee that the loop update will never result in
-	// `block==nil`.
+	// `block==nil` during block building. Verification, however, MUST protect
+	// against malformed input so performs a check.
 	for block := parent; ; block = block.ParentBlock() {
+		if block == nil {
+			return nil, false, fmt.Errorf("%w: settling at %d with parent %#x (%v)", errIncompleteBlockHistory, settleAt, parent.Hash(), parent.Number())
+		}
```
StephenButtolph: Removing this check does not cause any tests to fail. If this is actually needed (which I don't think it is), we should definitely add a regression test for it.

sae/blocks.go
```diff
@@ -5,42 +5,240 @@ package sae
...
+
+// ParseBlock parses the buffer as [rlp] encoding of a [types.Block]. It does
+// NOT populate the block ancestry, which is done by [VM.VerifyBlock] i.f.f.
+// verification passes.
+func (vm *VM) ParseBlock(ctx context.Context, buf []byte) (*blocks.Block, error) {
+	b := new(types.Block)
+	if err := rlp.DecodeBytes(buf, b); err != nil {
+		return nil, fmt.Errorf("rlp.DecodeBytes(..., %T): %v", b, err)
+	}
+	return vm.newBlock(b, nil, nil)
```
StephenButtolph: This seems to break the documented invariant on `ancestry`:
> // Invariant: ancestry is non-nil and contains non-nil pointers i.f.f. the
	// block hasn't itself been settled. A synchronous block (e.g. SAE genesis
	// or the last pre-SAE block) is always considered settled.

This block hasn't been settled, so it must contain non-nil pointers.

Of course, this can't actually be guaranteed here... So I think we need to update these invariants (or make some bigger change).

sae/blocks.go
```diff
@@ -5,42 +5,240 @@ package sae
...
+		return nil, errExecutionLagging
+	}
+
+	// Although, at the time of writing this, the implementation of
+	// [blocks.LastToSettleAt] guarantees that the returned block has finished
+	// execution, this isn't a stated invariant and it's possible for it to be
+	// more aggressive in the future.
+	wCtx, cancel := context.WithTimeout(ctx, 10*time.Millisecond)
+	defer cancel()
+	if err := lastSettled.WaitUntilExecuted(wCtx); err != nil {
```
StephenButtolph: Even though 10ms isn't very long at all, I don't think we should be waiting here (on the consensus thread). It seems like we should instead be checking `lastSettled.Executed()` and immediately erroring or progressing.

If we are actually hitting this wait, we probably either have a bug in `WaitForEvent` or need to rethink how we select the building timestamp.

sae/blocks.go
```diff
@@ -5,42 +5,240 @@ package sae
...
+			return nil, err
+		}
+	}
+
+	hdr := &types.Header{
+		ParentHash: parent.Hash(),
+		Number:     new(big.Int).Add(parent.Number(), common.Big1),
+		Time:       unix(blockTime),
+		Root:       lastSettled.PostExecutionStateRoot(),
+	}
```
StephenButtolph: Probably not an issue for this PR, but this doesn't expose sufficient functionality to do things like populating the ms timestamps, or updating the gas target, ect.

sae/http.go
```diff
@@ -10,13 +10,19 @@ import (
...
+	}, nil
 }
 
 // NewHTTPHandler returns the HTTP handler that will be invoked if a client
 // passes this VM's chain ID via the routing header described in the [common.VM]
 // documentation for this method.
 func (vm *VM) NewHTTPHandler(context.Context) (http.Handler, error) {
-	return nil, errUnimplemented
+	return vm.ethRPCServer()
 }
```
StephenButtolph: Unless we are trying to support H2 (we aren't here) I don't see a good reason to expose this

sae/http.go
```diff
@@ -10,13 +10,19 @@ import (
...
 // node, keyed by extension.
-func (vm *VM) CreateHandlers(context.Context) (map[string]http.Handler, error) {
-	return nil, errUnimplemented
+func (vm *VM) CreateHandlers(ctx context.Context) (map[string]http.Handler, error) {
+	s, err := vm.NewHTTPHandler(ctx)
+	if err != nil {
+		return nil, err
+	}
+	return map[string]http.Handler{
+		"/": s,
```
StephenButtolph: In subnet-evm and coreth we route the simple HTTP traffic to `/rpc` and the websocket traffic to `/ws`. We should do that here as well imo

sae/sae.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+	u, overflow := uint256.FromBig(b)
+	if overflow {
+		return nil, fmt.Errorf("big.Int %v overflows 256 bits", b)
+	}
+	return u, nil
+}
+
+type sMap[K comparable, V any] struct {
+	m sync.Map
```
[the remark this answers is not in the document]
ARR4N: I've changed it because it also removes the need for the type assertion.

sae/blocks.go
```diff
@@ -5,42 +5,240 @@ package sae
...
+		}
+
+		feeCap, err := uint256FromBig(tx.GasFeeCap(), errOnNil)
+		if err != nil {
+			return fmt.Errorf("tx %#x fee cap: %v", tx.Hash(), err)
+		}
+		tipCap, err := uint256FromBig(tx.GasTipCap(), errOnNil)
+		if err != nil {
+			return fmt.Errorf("tx %#x tip cap: %v", tx.Hash(), err)
+		}
```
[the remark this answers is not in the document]
ARR4N: Confirmed impossible, but I've left the check in place because it's trivial and already in a function that returns an error. Also see my [reply here](https://github.com/ava-labs/strevm/pull/48#discussion_r2630727400).

sae/blocks.go
```diff
@@ -5,42 +5,241 @@ package sae
...
+// ParseBlock parses the buffer as [rlp] encoding of a [types.Block]. It does
+// NOT populate the block ancestry, which is done by [VM.VerifyBlock] i.f.f.
+// verification passes.
+func (vm *VM) ParseBlock(ctx context.Context, buf []byte) (*blocks.Block, error) {
+	b := new(types.Block)
+	if err := rlp.DecodeBytes(buf, b); err != nil {
+		return nil, fmt.Errorf("rlp.DecodeBytes(..., %T): %v", b, err)
+	}
+	return vm.newBlock(b, nil, nil)
 }
```
StephenButtolph: In addition to the comments about `.Timestamp()` and `.Height()` functions in `VerifyBlock`, the consensus code is also vulnerable to misuse of those functions here.

It seems that `Parse` should be verifying that both `b.Time` and `b.Number` have reasonable values before returning here.

sae/blocks.go
```diff
@@ -5,42 +5,241 @@ package sae
...
+		var lim gas.Gas
+		for _, tx := range b.Transactions() {
+			lim += gas.Gas(tx.Gas())
+		}
+		if err := clock.AfterBlock(lim, vm.hooks, b.Header()); err != nil {
+			return nil, err
+		}
+	}
+
+	hdr := &types.Header{
```
StephenButtolph: Probably not super actionable right now, but figured I'd comment it since it came up in my review.
- `Coinbase` on C-chain and subnet-evm is typically set to the Blackhole address `common.Address{1}`. changing that here would change how fee burning would need to be calculated externally, so we probably want to keep that the same.
- `Difficulty` on C-chain and subnet-evm we set this explicitly to `1`. I'm not sure that it really matters... But probably makes sense not to change that pattern when SAE activates.
- All the upgrade values should be populated (if the upgrade calls for them), `BlobGasUsed`, `ExcessBlobGas`, `ParentBeaconRoot`. `Withdrawals` doesn't matter because we override the serialization not to include them ever.

sae/vm.go
```diff
@@ -1,28 +1,180 @@
...
 // then returns the message to send.
-func (vm *VM) WaitForEvent(ctx context.Context) (common.Message, error) {
-	return 0, errUnimplemented
+func (vm *VM) WaitForEvent(ctx context.Context) (snowcommon.Message, error) {
+	select {
+	case _, ok := <-vm.newTxs:
+		if !ok {
+			return 0, errors.New("VM closed")
+		}
+		return snowcommon.PendingTxs, nil
```
StephenButtolph: This implementation is really not sufficient. Specifically, returning `snowcommon.PendingTxs` does not guarantee that `BuildBlock` will be called.

Additionally, it may have been possible that `BuildBlock` was called but didn't include all txs from the mempool (which would result in txs stuck in the mempool).

Should there be a TODO added here?

Other things to consider: The mempool may not have actually finished reorging when this function is called, so we should probably wait for any outstanding reorgs to finish.

StephenButtolph (commented):
I'm not sure how polished the goal of this PR is. I wrote up concerns that I think could be fairly easily fixed... But it generally lgtm.

sae/vm_test.go
```diff
@@ -0,0 +1,270 @@
...
+			header: &types.Header{
+				Number: big.NewInt(1),
+				Time:   now + maxBlockFutureSeconds + 1,
+			},
+			wantErr: errBlockTooFarInFuture,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run("", func(t *testing.T) {
```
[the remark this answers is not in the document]
ARR4N: Whoops, that wasn't on purpose. I've named them Stephen and Gary.

sae/vm_test.go
```diff
@@ -0,0 +1,270 @@
...
+	rawVM   *VM
+	genesis *blocks.Block
+	wallet  *saetest.Wallet
+}
+
+func newSUT(tb testing.TB, numAccounts uint) (context.Context, *SUT) {
+	tb.Helper()
+
+	mempoolConf := legacypool.DefaultConfig // copies
+	mempoolConf.Journal = filepath.Join(tb.TempDir(), "transactions.rlp")
```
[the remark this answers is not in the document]
ARR4N: No idea. I've changed it to `/dev/null` (primarily to demonstrate that it's possible) because the only intent behind the `TempDir()` was to avoid interdependence between test runs.

sae/vm.go
```diff
@@ -1,28 +1,202 @@
...
+// Init initializes the [VM], similarly to the `Initialize` method of
+// [snowcommon.VM].
+func (vm *VM) Init(
+	snowCtx *snow.Context,
+	hooks hook.Points,
+	chainConfig *params.ChainConfig,
+	db ethdb.Database,
+	triedbConfig *triedb.Config,
+	lastSynchronous *blocks.Block,
+) error {
```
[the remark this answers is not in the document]
ARR4N: I've added a TODO.

PR #51, opened 2025-12-22
Add `BuildBlock` and `BlockRebuilderFrom` hooks

StephenButtolph:
This PR provides the minimal hooks for allowing block creation within the hooks. The `BuildBlock` function on the `BlockBuilder` interface will be expanded to take in more arguments in the future. This PR is intended to mainly introduce the shape of the block building hooks.

### DIFF
--- a/hook/hook.go
+++ b/hook/hook.go
@@ -24,7 +24,19 @@ import (
 )
 
 // Points define user-injected hook points.
+//
+// Directly using this interface as a [BlockBuilder] is indicative of this node
+// locally building a block. Calling [Points.BlockRebuilderFrom] with an
+// existing block is indicative of this node reconstructing a block built
+// elsewhere during verification.
 type Points interface {
+	BlockBuilder
+	// BlockRebuilderFrom returns a [BlockBuilder] that will attempt to
+	// reconstruct the provided block. If the provided block is valid for
+	// inclusion, then the returned builder MUST be able to reconstruct an
+	// identical block.
+	BlockRebuilderFrom(block *types.Block) BlockBuilder
+
 	// GasTargetAfter returns the gas target that should go into effect
 	// immediately after the provided block.
 	GasTargetAfter(*types.Header) gas.Gas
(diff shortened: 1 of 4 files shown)


hook/hook.go
```diff
@@ -24,7 +24,18 @@ import (
...
+// Directly using this interface as a [BlockBuilder] is indicative of this node
+// locally building a block. Calling [Points.BlockBuilder] with an existing
+// block is indicative of this node reconstructing a block built elsewhere
+// during verification.
 type Points interface {
+	BlockBuilder
+	// BlockBuilder returns a [BlockBuilder] that will attempt to reconstruct
+	// the provided block. If the provided block is valid for inclusion, then
+	// the returned builder MUST be able to reconstruct an identical block.
+	BlockBuilder(block *types.Block) BlockBuilder
```
```suggestion
	BlockRebuilderFrom(block *types.Block) BlockBuilder
```

Or just `BlockRebuilder()` although I like how the `From` is very suggestive of intended usage. You'll also need to update the comment on `Points`.

hook/hookstest/stub.go
```diff
@@ -22,6 +23,20 @@ type Stub struct {
 
 var _ hook.Points = (*Stub)(nil)
 
+// NewBlock calls [types.NewBlock] with its arguments.
+func (*Stub) NewBlock(
+	header *types.Header,
+	txs []*types.Transaction,
+	receipts []*types.Receipt,
+) *types.Block {
+	return types.NewBlock(header, txs, nil, receipts, trie.NewStackTrie(nil))
```
ARR4N:
```suggestion
	return types.NewBlock(header, txs, nil, receipts, saetest.TrieHasher())
```

Although it's semantically identical, it stops the reader having to think about the appropriateness of `trie.NewStackTrie(nil)` here because it's really weird IMO.

sae/blocks.go
```diff
@@ -75,14 +76,28 @@ func (vm *VM) BuildBlock(ctx context.Context, bCtx *block.Context) (*blocks.Bloc
...
 // buildBlock implements the block-building logic shared by [VM.BuildBlock] and
 // [VM.VerifyBlock].
-func (vm *VM) buildBlock(ctx context.Context, bCtx *block.Context, parent *blocks.Block, blockTime time.Time, candidates []*txgossip.LazyTransaction) (*blocks.Block, error) {
+func (vm *VM) buildBlock(
+	ctx context.Context,
+	bCtx *block.Context,
+	parent *blocks.Block,
+	blockTime time.Time,
+	candidates []*txgossip.LazyTransaction,
+	builder hook.BlockBuilder,
```
ARR4N: Nice :) I like the simplicity of this PR.

ARR4N (approved):

PR #53, opened 2025-12-23
Use `worstcase.State` in `VM.buildBlock`

StephenButtolph:
This actually enforces worst-case state checks for the new blocks + blocks being verified.

It additionally adds more verification of the provided block time during verification to prevent bad timestamps.

### DIFF
--- a/blocks/settlement.go
+++ b/blocks/settlement.go
@@ -142,9 +142,19 @@ func (b *Block) Settles() []*Block {
 //
 // If the two arguments are the same block, Range returns an empty slice.
 func Range(start, end *Block) []*Block {
-	var chain []*Block
-	for b := end; b.Hash() != start.Hash(); b = b.ParentBlock() {
-		chain = append(chain, b)
+	startHeight := start.Height()
+	endHeight := end.Height()
+	if endHeight <= startHeight {
+		return nil
+	}
+
+	var (
+		chain = make([]*Block, endHeight-startHeight)
+		b     = end
+	)
+	for i := range chain {
+		chain[i] = b
+		b = b.ParentBlock()
 	}
 	slices.Reverse(chain)
 	return chain


sae/vm_test.go
```diff
@@ -270,3 +274,94 @@ func TestSyntacticBlockChecks(t *testing.T) {
...
+			)
+			b := blockstest.NewBlock(
+				t,
+				ethB,
+				nil,
+				nil,
+			)
+			snowB, err := sut.ParseBlock(ctx, b.Bytes())
+			require.NoErrorf(t, err, "ParseBlock(...)")
+			require.ErrorIs(t, snowB.Verify(t.Context()), tt.wantErr, "Verify()")
```
StephenButtolph: The alternative was:
```suggestion
			err := sut.rawVM.VerifyBlock(t.Context(), nil, b)
			require.ErrorIs(t, err, tt.wantErr, "VerifyBlock()")
```
But I wasn't sure if that was preferred since it wasn't exposed as a method on `sut`.

sae/blocks.go
```diff
@@ -104,74 +101,148 @@ func (vm *VM) buildBlock(
...
+		return nil, fmt.Errorf("%w: %s > %s", errBlockTimeAfterMaximum, blockTime, maxBlockTime)
+	}
+
+	settleAt := unix(blockTime.Add(-saeparams.Tau))
 	lastSettled, ok, err := blocks.LastToSettleAt(settleAt, parent)
 	if err != nil {
 		return nil, err
 	}
 	if !ok {
-		log.Warn("Execution lagging when determining last block to settle")
```
ARR4N: We should know when this happens because it suggests that the gas target is too high.

sae/blocks.go
```diff
@@ -104,74 +101,148 @@ func (vm *VM) buildBlock(
...
-		return nil, fmt.Errorf("state.New([post-execution state root %#x of block %d], ...): %v", b.PostExecutionStateRoot(), b.Height(), err)
-	}
-
-	// TODO(arr4n) worst-case tx validation will happen here. This is just a
-	// sketch of the process.
-	history := parent.WhenChildSettles(lastSettled)
-	var receipts types.Receipts
-	for _, b := range history {
-		receipts = append(receipts, b.Receipts()...)
+		log.Warn("Settled state not available")
```
ARR4N: Is this in lieu of the `if !lastSettled.Executed()` check?

ARR4N:
```suggestion
		log.Warn(
			"Settled state not available",
			zap.Error(err),
		)
```

I think it's worth including the error so the poor on-call person seeing this in the logs at least has something to work with. I think the `worstcase.errSettledBlockNotExecuted` message should also be updated to "block for settling not finished execution yet" to make it obvious that this is a transient state.

sae/blocks.go
```diff
@@ -104,74 +101,148 @@ func (vm *VM) buildBlock(
...
+			break
+		}
+
 		tx, ok := ltx.Resolve()
 		if !ok {
 			continue
 		}
-		_ = sdb // balance checks
+
+		if err := state.ApplyTx(tx); err != nil {
```
ARR4N: Please log these errors at `DEBUG`. Silent non-inclusion of transactions was the hardest thing to debug in the PoC.

sae/blocks.go
```diff
@@ -104,74 +101,148 @@ func (vm *VM) buildBlock(
...
+			continue
+		}
 		included = append(included, tx)
 	}
 
+	// TODO: Should the [hook.BlockBuilder] populate [types.Header.GasUsed] so
+	// that [hook.Op.Gas] can be included?
```
ARR4N: Please explain. How is `hook.Op.Gas` not included if you're using `state.GasUsed()`, which has had the ops applied?

sae/blocks.go
```diff
@@ -104,74 +101,148 @@ func (vm *VM) buildBlock(
...
+		included []*types.Transaction
+	)
 	for _, ltx := range candidates {
+		// If we don't have enough gas remaining in the block for the minimum
+		// gas amount, we are done including transactions.
+		if remainingGas := state.GasLimit() - state.GasUsed(); remainingGas < params.TxGas {
+			break
+		}
+
 		tx, ok := ltx.Resolve()
```
ARR4N: Thoughts on always resolving? I don't know much about the history of the `geth` mempool and why the lazy transaction was introduced. We could have `worstcase.State` accept a `txgossip.LazyTransaction` as it includes the sender (i.e. no more need for a signer) and `txpool.LazyTransaction` includes enough to convert it to a `hook.Op` without calling `Resolve()`.

sae/blocks.go
StephenButtolph: I removed this because malicious nodes can force this to happen (and generally we don't log at a default enabled level to prevent malicious log spam).

But we do have a precedent for logging errors like this _after_ filtering so only the block producer could cause a log spam... So I think this is fine as a Warn.

sae/blocks.go
StephenButtolph: The `.Executed()` check is performed internally (as that is where the execution artifacts are used).

I changed the log to refer to the worst-case state rather than the settled state, since that follows more from the actual code (and doesn't require knowledge of how worstcase is initialized)

sae/blocks.go
StephenButtolph: I'm also not particularly familiar with `LazyTransaction` handling.

While I don't think it would immediately be a problem, I don't love accepting a LazyTransaction in the worstcase logic. It seems pretty fragile to be using a signer that isn't clearly guaranteed to be the same between the execution logic and the worstcase logic.

We'll need to resolve anyways if we actually include it in the block.

sae/blocks.go
```diff
@@ -180,6 +251,32 @@ func (vm *VM) buildBlock(
 	return vm.newBlock(ethB, parent, lastSettled)
 }
 
+// unsettledAncestry returns the ancestry of blocks from `parent` (inclusive) to
+// the (non-inclusive) `settledHeight`; in order of oldest to newest.
+//
+// It is assumed that `parent` was successfully verified and has not been
+// rejected.
+func unsettledAncestry(parent *blocks.Block, settledHeight uint64) []*blocks.Block {
```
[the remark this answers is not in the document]
StephenButtolph: Replaced `WhenChildSettles` with `blocks.Range`

sae/vm_test.go
ARR4N: I deliberately do everything via the methods that consensus will see so that we get a full integration test, including of the adaptor.

blocks/settlement.go
```diff
@@ -130,45 +130,24 @@ func (b *Block) Settles() []*Block {
...
-// arguments have the same block hash.
-func settling(lastOfParent, lastOfCurr *Block) []*Block {
-	var settling []*Block
-	// TODO(arr4n) abstract this to combine functionality with iterators
-	// introduced by @StephenButtolph.
-	for s := lastOfCurr; s.Hash() != lastOfParent.Hash(); s = s.ParentBlock() {
-		settling = append(settling, s)
+// If the two arguments are the same block, it returns an empty slice.
+func Range(start, end *Block) []*Block {
+	var chain []*Block
```
ARR4N: (optional) You're the benchmarking guru so I'll leave it up to you to decide if you want to specify the length not use `append()`.

blocks/settlement_test.go
```diff
@@ -36,7 +36,10 @@ func ExampleBlock_WhenChildSettles() {
 
 	// Returns the (possibly empty) slice of blocks that would be settled by the
 	// block being built.
-	_ = parent.WhenChildSettles(settle)
+	_ = Range(parent.LastSettled(), settle)
+	// Returns the (possibly empty) slice of blocks that would be left unsettled
+	// by the block being built.
+	_ = Range(settle, parent)
```
ARR4N: Nice :)

I was thinking through the arguments after we chatted and realised it would fall out really elegantly, especially when also looking inside `Block.Settles()`.

blocks/settlement_test.go
```diff
@@ -163,26 +166,26 @@ func TestSettles(t *testing.T) {
 	for _, b := range blocks[1:] {
 		tests = append(tests, testCase{
 			name: fmt.Sprintf("Block(%d).WhenChildSettles([same as parent])", b.Height()),
```
ARR4N:
```suggestion
			name: "Range([identical blocks])",
```

sae/blocks.go
```diff
@@ -131,11 +131,13 @@ func (vm *VM) buildBlock(
...
 	if err != nil {
-		log.Warn("Settled state not available")
+		log.Warn("Worst case state not able to be created",
+			zap.Error(err),
+		)
 		return nil, err
 	}
 
-	for _, b := range unsettledAncestry(parent, lastSettled.Height()) {
+	for _, b := range blocks.Range(lastSettled, parent) {
```
ARR4N:
```suggestion
	unsettled := blocks.Range(lastSettled, parent)
	for _, b := range unsettled {
```

As much as `blocks.Range()` simplifies everything, I think it's still useful to communicate the general intent, and will recommend the same below. Without this, we would force the reader to figure out the implication of the arguments to `Range()` whereas now they just need to confirm the correctness (or take it for granted).

sae/blocks.go
```diff
@@ -239,7 +249,7 @@ func (vm *VM) buildBlock(
 	}
 
 	var receipts types.Receipts
-	for _, b := range parent.WhenChildSettles(lastSettled) {
+	for _, b := range blocks.Range(parent.LastSettled(), lastSettled) {
```
ARR4N:
```suggestion
	settling := blocks.Range(parent.LastSettled(), lastSettled)
	for _, b := range settling {
```

See rationale above.

sae/blocks.go
```diff
@@ -114,64 +124,135 @@ func (vm *VM) buildBlock(
...
-		var lim gas.Gas
-		for _, tx := range b.Transactions() {
-			lim += gas.Gas(tx.Gas())
+	for _, b := range blocks.Range(lastSettled, parent) {
+		log := log.With(
+			zap.Uint64("block_height", b.Height()),
+			zap.Stringer("block_hash", b.Hash()),
+		)
+		if err := state.StartBlock(b.Header()); err != nil {
+			log.Warn("Could not start historical worst case calculation",
```
ARR4N:
```suggestion
			log.Warn("Could not start historical worst-case calculation",
```

sae/blocks.go
```diff
@@ -114,64 +124,135 @@ func (vm *VM) buildBlock(
...
 	}
 
+	// TODO: Should the [hook.BlockBuilder] populate [types.Header.GasUsed] so
+	// that [hook.Op.Gas] can be included?
+	hdr.GasUsed = state.GasUsed()
+
+	// Although we never interact with the worst-case state after this point, we
+	// still mark the block as finished to align with normal execution.
+	if err := state.FinishBlock(); err != nil {
+		log.Warn("Could not finish worst case block calculation",
```
ARR4N:
```suggestion
		log.Warn("Could not finish worst-case block calculation",
```

sae/blocks.go
```diff
@@ -193,14 +280,12 @@ func (vm *VM) VerifyBlock(ctx context.Context, bCtx *block.Context, b *blocks.Bl
...
 	}
 
-	switch height, accepted := b.Height(), vm.lastAccepted.Load().Height(); {
-	case height != parent.Height()+1:
-		return fmt.Errorf("non-incrementing block height; verifying at %d with parent at %d", height, parent.Height())
-	case height <= accepted:
-		return fmt.Errorf("verifying block at height %d <= last-accepted (%d)", height, accepted)
+	// Sanity check that we aren't verifying an accepted block.
+	if height, accepted := b.Height(), vm.lastAccepted.Load().Height(); height <= accepted {
+		return fmt.Errorf("%w at height %d <= last-accepted (%d)", errFinalizedBlock, height, accepted)
```
ARR4N: What do you mean by "finalized block"?

sae/vm_test.go
```diff
@@ -270,3 +274,94 @@ func TestSyntacticBlockChecks(t *testing.T) {
...
+				nil,
+				tt.receipts,
+				saetest.TrieHasher(),
+			)
+			b := blockstest.NewBlock(
+				t,
+				ethB,
+				nil,
+				nil,
+			)
```
```suggestion
			b := blockstest.NewBlock(t, ethB, nil, nil)
```

sae/vm_test.go
```diff
@@ -270,3 +274,94 @@ func TestSyntacticBlockChecks(t *testing.T) {
...
+	lastAcceptedID, err := sut.LastAccepted(ctx)
+	require.NoError(t, err, "LastAccepted()")
+	lastAccepted, err := sut.GetBlock(ctx, lastAcceptedID)
+	require.NoError(t, err, "GetBlock(lastAcceptedID)")
+
+	tests := []struct {
+		name           string
+		parentHash     common.Hash
+		acceptedHeight bool
+		time           uint64
```
```suggestion
		time           uint64 // defaults to `now` if zero
```

sae/vm_test.go
```diff
@@ -270,3 +274,94 @@ func TestSyntacticBlockChecks(t *testing.T) {
...
+			wantErr:    errUnknownParent,
+		},
+		{
+			name:           "already_finalized",
+			acceptedHeight: true,
+			wantErr:        errFinalizedBlock,
+		},
+		{
+			name:    "block_time_under_minimum",
+			time:    1,
```
```suggestion
			time:    saeparams.TauSeconds-1,
```

(readability) It explains why to the reader.

ARR4N (approved):
Everything is about readability and nothing requires re-review unless you disagree with something and you think I'll care.

sae/blocks.go
```diff
@@ -64,29 +63,26 @@ func (vm *VM) ParseBlock(ctx context.Context, buf []byte) (*blocks.Block, error)
...
-		vm.mempool.TransactionsByPriority(filter),
+		vm.mempool.TransactionsByPriority,
 		vm.hooks,
 	)
 }
 
-var errExecutionLagging = errors.New("execution lagging for settlement")
+// Max time from current time allowed for blocks, before they're considered
+// future blocks and fail verification.
+const maxFutureBlockTime = 10 * time.Second
```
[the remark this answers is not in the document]
StephenButtolph: A fair question... This isn't as easy as it would seem because the P-chain actually didn't include timestamps in the blocks at network launch... So the invariants around `block.Timestamp` are a bit subtle in the consensus code.

PR #55, opened 2025-12-24
Remove references to non-existent function

StephenButtolph:
`Block.RestorePostExecutionState` doesn't exist (and neither does `Block.ResotrePostExecutionState`)


ARR4N (approved):

ARR4N:
These might be re-introduced later during database recovery. See my PoC branch if interested.

PR #58, opened 2025-12-29
Allow hooks to influence block building

StephenButtolph:
1. `BlockEvents` is added to allow the hooks to enforce the ACP-226 minimum block delay.
2. `WaitForEvent` is added to allow the hooks to signal that an atomic tx is in the atomic mempool and that a block should be built (even if there are no pending eth txs)

### DIFF
--- a/blocks/blockstest/blocks.go
+++ b/blocks/blockstest/blocks.go
@@ -110,8 +110,9 @@ func NewGenesis(tb testing.TB, db ethdb.Database, config *params.ChainConfig, al
 	options.ApplyTo(conf, opts...)
 
 	gen := &core.Genesis{
-		Config: config,
-		Alloc:  alloc,
+		Config:    config,
+		Timestamp: conf.timestamp,
+		Alloc:     alloc,
 	}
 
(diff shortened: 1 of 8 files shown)


sae/vm.go
```diff
@@ -167,6 +167,35 @@ func (vm *VM) signalNewTxsToEngine() {
...
+	handleEvent := func(f func(context.Context) (snowcommon.Message, error)) {
+		defer cancel()
+		msg, err := f(ctx)
+		events <- event{msg, err}
+	}
+	go handleEvent(vm.waitForEvent)
+	go handleEvent(vm.hooks.WaitForEvent)
+	e := <-events
+	return e.msg, e.err
+}
```
StephenButtolph: Technically, this function could be implemented by _wrapping_ the sae VM rather than providing its own hook. Although that would change the invariant on `BuildBlock`, as we would no longer only be calling BuildBlock in response to a message returned from this VM.

I'm happy with either approach, so if you (@ARR4N) would prefer to keep the hook interface minimal, we could close this PR and instead just have coreth wrap this method.

PR #59, opened 2025-12-30
feat: `VM.AcceptBlock()` implementation w/ invariant documentation

ARR4N:
The majority of this PR is documentation that defines the order in which side effects of the block lifecycle MUST occur. The implementation of `VM.AcceptBlock()` (which, by definition, includes block settlement) necessitated these ordering invariants to avoid race conditions in current and future code.

As the block lifecycle is now (roughly) complete, I thought the most valuable test would be integration rather than unit. The PR therefore includes the necessary (and concise) RPC-backend methods to support `eth_getHeaderByNumber` as it exercises block labels as well as canonical blocks.

### DIFF
--- a/sae/rpc.go
+++ b/sae/rpc.go
@@ -5,6 +5,7 @@ package sae
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"math/big"
 
@@ -64,7 +65,9 @@ func (b *ethAPIBackend) GetTd(context.Context, common.Hash) *big.Int {
 
 func (b *ethAPIBackend) BlockByNumber(ctx context.Context, n rpc.BlockNumber) (*types.Block, error) {
 	num, err := b.resolveBlockNumber(n)
-	if err != nil {
+	if errors.Is(err, errFutureBlockNotResolved) {
+		return nil, nil
+	} else if err != nil {
 		return nil, err
 	}
 	return rawdb.ReadBlock(
@@ -73,6 +76,8 @@ func (b *ethAPIBackend) BlockByNumber(ctx context.Context, n rpc.BlockNumber) (*
 		num,
 	), nil
 }
+
+var errFutureBlockNotResolved = errors.New("not accepted yet")
 
 func (b *ethAPIBackend) resolveBlockNumber(bn rpc.BlockNumber) (uint64, error) {
 	head := b.vm.last.accepted.Load().Height()
@@ -92,7 +97,7 @@ func (b *ethAPIBackend) resolveBlockNumber(bn rpc.BlockNumber) (uint64, error) {
 	}
 	n := uint64(bn) //nolint:gosec // Non-negative check performed above
 	if n > head {
-		return 0, fmt.Errorf("block %d not accepted yet", n)
+		return 0, fmt.Errorf("%w: block %d", errFutureBlockNotResolved, n)
 	}
 	return n, nil
 }


sae/blocks.go
```diff
@@ -115,7 +115,7 @@ func (vm *VM) buildBlock(
 	if parentTime := parent.BuildTime(); hdr.Time < parentTime {
 		return nil, fmt.Errorf("%w: %d < %d", errBlockTimeBeforeParent, hdr.Time, parentTime)
 	}
-	if maxTime := uint64(vm.config.Now().Add(maxFutureBlockTime).Unix()); hdr.Time > maxTime { //nolint:gosec // Time won't overflow for quite a while
+	if maxTime := unix(vm.config.Now().Add(maxFutureBlockTime)); hdr.Time > maxTime {
```
ARR4N: (fyi) I implemented the `unix()` function specifically to avoid the need for `//nolint:gosec` everywhere. Sorry for not picking it up in my review of your PR.

sae/vm_test.go
```diff
@@ -291,9 +385,67 @@ func TestSyntacticBlockChecks(t *testing.T) {
...
+	// Causes [VM.AcceptBlock] to wait until the block has executed.
+	require.NoError(t, sut.SetState(ctx, snow.Bootstrapping), "SetState(Bootstrapping)")
+
+	unsettled := []*blocks.Block{sut.genesis}
+	last := func() *blocks.Block {
+		return unsettled[len(unsettled)-1]
+	}
+	sut.genesis = nil // allow it to be GCd when appropriate
+
+	rng := rand.New(rand.NewPCG(0, 0)) //nolint:gosec // Reproducibility is useful for tests
```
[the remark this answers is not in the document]
ARR4N: The recommended change is a "what" comment, which can be inferred from the code itself. Leaving the "why" comment is sufficient.

ARR4N:
As usual, Copilot review was pointless. It's a glorified spell check that rehashed my PR description and made some questionable suggestions to make comments unnecessarily verbose.

sae/consensus.go
```diff
@@ -32,15 +36,89 @@ func (vm *VM) SetPreference(ctx context.Context, id ids.ID, bCtx *block.Context)
...
+	{
+		batch := vm.db.NewBatch()
+
+		// D(s \in S); yes, I know it's in a batch
+		if s := settles; len(s) > 0 {
+			// Note that while we treat SAFE and FINALIZED identically, there is
+			// no notion of the former in [rawdb]. Furthermore, the LAST label
+			// is reserved for the last executed so it too isn't persisted here.
+			rawdb.WriteFinalizedBlockHash(batch, s[len(s)-1].Hash())
+		}
```
StephenButtolph: While slightly less efficient, I feel like it's more clear
```suggestion
		if len(settles) > 0 {
			// Note that while we treat SAFE and FINALIZED identically, there is
			// no notion of the former in [rawdb]. Furthermore, the LAST label
			// is reserved for the last executed so it too isn't persisted here.
			rawdb.WriteFinalizedBlockHash(batch, b.LastSettled().Hash())
		}
```
Technically, we could even remove the `if` and always write the finalized block hash, but reducing DB IO is always a good thing (even at the slight reduction in readability).

sae/consensus.go
```diff
@@ -32,15 +36,89 @@ func (vm *VM) SetPreference(ctx context.Context, id ids.ID, bCtx *block.Context)
...
-	return vm.exec.Enqueue(ctx, b)
+	// Recall the terminology and ordering from the invariants document:
+	// - (D)isk then (M)emory then (I)nternal then e(X)ternal.
+	// - B accepted after all of B.Settles() settled
+
+	settles := b.Settles()
+	{
+		batch := vm.db.NewBatch()
+
+		// D(s \in S); yes, I know it's in a batch
```
StephenButtolph: IDEs don't render latex inline, so if we want to use the mathematical notation here, I think we should be using the unicode symbols. If we don't want to use the unicode symbols, then I think we should just write it out with words.
```suggestion
		// D(s ∈ S); yes, I know it's in a batch
```
(This suggestion applies to a number of places, I'll just leave it here for you to decide how best to proceed).

sae/consensus.go
ARR4N: > While slightly less efficient, I feel like it's more clear
💯

StephenButtolph (approved):

PR #60, opened 2025-12-31
Introduce p2p gossip

StephenButtolph:
This PR wires the avalanchego gossip package into the SAE VM to gossip txs over the p2p network.

The unit test is trash, but I'm not sure how easily we can clean it up. Talking with Arran, it might make sense to rely primarily on e2e tests for this, but to keep the test as-is for a best-effort quick unit test.

### DIFF
--- a/sae/vm_test.go
+++ b/sae/vm_test.go
@@ -678,7 +678,7 @@ func newNetworkedSUTs(tb testing.TB, numValidators, numNonValidators int) *netwo
 
 		// Non-validators do not connect to other non-validators.
 		var nonValidatorPeers map[ids.NodeID]*SUT
-		if _, ok := validatorNodes[selfID]; !ok {
+		if _, ok := validatorNodes[selfID]; ok {
 			nonValidatorPeers = nonValidatorNodes
 		}
 		maps.Copy(peers, nonValidatorPeers)


sae/vm_test.go
```diff
@@ -515,3 +537,195 @@ func TestSemanticBlockChecks(t *testing.T) {
...
+
+func requireNotReceiveTx(tb testing.TB, nodes []*SUT, txHash common.Hash) {
+	tb.Helper()
+	for _, sut := range nodes {
+		assert.False(tb, sut.rawVM.mempool.Has(ids.ID(txHash)), "tx %x was gossiped to node %s", txHash, sut.rawVM.snowCtx.NodeID)
+	}
+	require.False(tb, tb.Failed())
+}
+
+func TestGossip(t *testing.T) {
```
StephenButtolph: It feels a bit crazy that we have like 200 lines of setup for such a small test, but hopefully we'll extend the usage of the `networkSUT` (or replace it with Josh's setup)

sae/vm_test.go
```diff
@@ -112,11 +124,18 @@ func newSUT(tb testing.TB, numAccounts uint, opts ...sutOption) (context.Context
 
 	logger := saetest.NewTBLogger(tb, logging.Warn)
 	ctx := logger.CancelOnError(tb.Context())
-	snowCtx := snowtest.Context(tb, ids.GenerateTestID())
+
+	snowCtx := snowtest.Context(tb, chainID)
+	validators := snowCtx.ValidatorState.(*validatorstest.State)
+	validators.GetCurrentHeightF = func(context.Context) (uint64, error) {
+		return 0, nil
+	}
```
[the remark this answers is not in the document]
StephenButtolph: Linked PR was merged and the PR was updated to depend on it.

sae/vm_test.go
```diff
@@ -515,3 +535,204 @@ func TestSemanticBlockChecks(t *testing.T) {
...
+		}
+	}
+	getValidatorSetF := func(context.Context, uint64, ids.ID) (map[ids.NodeID]*validators.GetValidatorOutput, error) {
+		return validatorSet, nil
+	}
+	for _, sut := range allNodes {
+		sut.validators.GetValidatorSetF = getValidatorSetF
+	}
+
+	for nodeID, sut := range allNodes {
```
ARR4N: I keep looking back up here to see what `nodeID` refers to; I think `selfID` will be clearer (for example, removing the need for the `// self-registration` comment).

sae/vm_test.go
```diff
@@ -515,3 +535,204 @@ func TestSemanticBlockChecks(t *testing.T) {
...
+		peers := make(map[ids.NodeID]*SUT, len(allNodes))
+		peers[nodeID] = sut // self-registration
+		maps.Copy(peers, validatorNodes)
+
+		// Non-validators do not connect to other non-validators.
+		var nonValidatorPeers map[ids.NodeID]*SUT
+		if _, isValidator := validatorNodes[nodeID]; isValidator {
+			nonValidatorPeers = nonValidatorNodes
+		}
```
```suggestion
		if _, ok := validatorNodes[selfID]; ok {
			nonValidatorPeers = nonValidatorNodes
		}
```

See comment above for rationale behind `selfID` over `nodeID`.

sae/vm_test.go
```diff
@@ -515,3 +535,204 @@ func TestSemanticBlockChecks(t *testing.T) {
...
+		maps.Copy(peers, nonValidatorPeers)
+
+		// All messages are delivered in a new goroutine to prevent re-entrant
+		// calls, which would result in a deadlock.
+		sender := sut.sender
+		sender.SendAppRequestF = func(ctx context.Context, s set.Set[ids.NodeID], r uint32, b []byte) error {
+			go func() {
+				tb.Helper()
```
ARR4N: I think in this case we want errors reported at their source so `tb.Helper()` is, umm, unhelpful.

sae/vm_test.go
```diff
@@ -515,3 +535,204 @@ func TestSemanticBlockChecks(t *testing.T) {
...
+		var nonValidatorPeers map[ids.NodeID]*SUT
+		if _, isValidator := validatorNodes[nodeID]; isValidator {
+			nonValidatorPeers = nonValidatorNodes
+		}
+		maps.Copy(peers, nonValidatorPeers)
+
+		// All messages are delivered in a new goroutine to prevent re-entrant
+		// calls, which would result in a deadlock.
+		sender := sut.sender
+		sender.SendAppRequestF = func(ctx context.Context, s set.Set[ids.NodeID], r uint32, b []byte) error {
```
ARR4N: `s`, `r`, and `b` are too terse for this context.

sae/vm.go
```diff
@@ -155,6 +160,57 @@ func (vm *VM) Init(
...
+		network, validatorPeers, err := newNetwork(snowCtx, sender, vm.metrics)
+		if err != nil {
+			return fmt.Errorf("newNetwork(...): %v", err)
+		}
+
+		systemConfig := gossip.SystemConfig{
+			Log:       snowCtx.Log,
+			Registry:  vm.metrics,
+			Namespace: "gossip",
+		}
```
StephenButtolph: There are a lot of options here that we may want to allow users to configure... But I think we can add configurations more as an "as needed" basis.

sae/vm_test.go
```diff
@@ -588,3 +612,204 @@ func TestSemanticBlockChecks(t *testing.T) {
 		})
 	}
 }
+
+type networkSUT struct {
```
```suggestion
type networkedSUTs struct {
```

(optional, pedantry) The original name suggests that it's the network that's under test when it's actually the individual `SUT` objects.

If you agree, then `newNetworkSUT()` should probably be `newNetworkedSUTs()` too.

sae/vm_test.go
```diff
@@ -588,3 +612,204 @@ func TestSemanticBlockChecks(t *testing.T) {
...
+		}
+	}
+
+	validatorSet := make(map[ids.NodeID]*validators.GetValidatorOutput)
+	for _, sut := range validatorNodes {
+		nodeID := sut.rawVM.snowCtx.NodeID
+		validatorSet[nodeID] = &validators.GetValidatorOutput{
+			NodeID:    nodeID,
+			PublicKey: sut.rawVM.snowCtx.PublicKey,
+			Weight:    1,
```
ARR4N: What is this?! Simplex? 😝

sae/vm_test.go
```diff
@@ -588,3 +612,204 @@ func TestSemanticBlockChecks(t *testing.T) {
...
+		sender := sut.sender
+		sender.SendAppRequestF = func(ctx context.Context, s set.Set[ids.NodeID], r uint32, b []byte) error {
+			go func() {
+				tb.Helper()
+
+				for peerID := range s {
+					if peer, ok := peers[peerID]; ok {
+						assert.NoErrorf(tb, peer.AppRequest(ctx, nodeID, r, mockable.MaxTime, b), "sending request to %s", peerID)
+					} else {
+						assert.NoErrorf(tb, sut.AppRequestFailed(ctx, peerID, r, snowcommon.ErrTimeout), "sending request failed to %s", peerID)
```
ARR4N: Why `ErrTimeout`?

```suggestion
						assert.NoErrorf(tb, sut.AppRequestFailed(ctx, peerID, r, snowcommon.ErrTimeout), "AppRequestFailed(ctx, %s, ...)", peerID)
```

"sending request failed" reads as if you failed to send the request. I considered suggesting "sending request-failed" but simply [identifying the function](https://google.github.io/styleguide/go/decisions#identify-the-function) fixes the problem.

Although it's not ambiguous, you might as well do it for `AppRequest()` too.

sae/vm_test.go
```diff
@@ -588,3 +612,204 @@ func TestSemanticBlockChecks(t *testing.T) {
...
+}
+
+func requireReceiveTx(tb testing.TB, nodes []*SUT, txHash common.Hash) {
+	tb.Helper()
+	for _, sut := range nodes {
+		assert.Eventuallyf(tb, func() bool {
+			return sut.rawVM.mempool.Has(ids.ID(txHash))
+		}, 5*time.Second, 100*time.Millisecond, "tx %x not gossiped to node %s", txHash, sut.rawVM.snowCtx.NodeID)
+	}
+	require.False(tb, tb.Failed())
```
ARR4N: Use `tb.FailNow()` instead (and below) otherwise there will be a strange "expected true to be false" in the errors.

sae/vm_test.go
```diff
@@ -588,3 +612,204 @@ func TestSemanticBlockChecks(t *testing.T) {
...
+		validators:    slices.Collect(maps.Values(validatorNodes)),
+		nonValidators: slices.Collect(maps.Values(nonValidatorNodes)),
+	}
+}
+
+func requireReceiveTx(tb testing.TB, nodes []*SUT, txHash common.Hash) {
+	tb.Helper()
+	for _, sut := range nodes {
+		assert.Eventuallyf(tb, func() bool {
+			return sut.rawVM.mempool.Has(ids.ID(txHash))
```
ARR4N: Should we call `Sync()` on the mempool first? Same below for the not-received helper.

sae/vm_test.go
```diff
@@ -588,3 +612,204 @@ func TestSemanticBlockChecks(t *testing.T) {
...
+			go func() {
+				tb.Helper()
+
+				for peerID := range c.NodeIDs {
+					if peer, ok := peers[peerID]; ok {
+						assert.NoErrorf(tb, peer.AppGossip(ctx, nodeID, b), "sending gossip to %s", peerID)
+					}
+				}
+				var sent set.Set[ids.NodeID]
+				sent.Union(c.NodeIDs)
```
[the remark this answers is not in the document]
StephenButtolph: yeah

sae/vm_test.go
StephenButtolph: Calling `Sync` doesn't seem to impact `Has`. Even if it did, calling `Sync` isn't sufficient, as the tx may have been introduced to the gossip system but not sent to the mempool yet.

So I'd argue that calling `Sync` would be more confusing than anything else.

sae/vm_test.go
StephenButtolph: `ErrTimeout` is the error returned by the p2p code. We don't currently distinguish between requests to people that we aren't connected to and people that don't respond. Both of those cases will return an `ErrTimeout` (although, to be clear, we don't actually wait for the timeout if we aren't connected to them).

PR #62, opened 2026-01-02
Fix `TestAcceptBlock` flake

StephenButtolph:
The root cause is that the test assumes that `Block.WaitUntilExecuted()` guarantees that `Executor.LastExecuted()` will include that block.

But: https://github.com/ava-labs/strevm/blob/4e465a5ef554daf6fa1d76ae259bc812a3a7689d/saexec/execution.go#L206-L209

`MarkExecuted` closes the channel to unblock `WaitUntilExecuted` before updating the atomic that reports `LastExecuted`.

The test was updated to wait for an `External Indicator` rather than depending on the order of `Internal Indicator`s.

### DIFF
--- a/sae/vm_test.go
+++ b/sae/vm_test.go
@@ -235,7 +235,7 @@ func (s *SUT) waitUntilExecuted(tb testing.TB, b *blocks.Block) {
 		case <-ctx.Done():
 			tb.Fatalf("waiting for block %d to execute: %v", b.Height(), ctx.Err())
 		case err := <-sub.Err():
-			tb.Fatalf("%T.SubscribeNewHead().Err() returned error: %v", s.Client, err)
+			tb.Fatalf("%T.SubscribeNewHead().Err() returned: %v", s.Client, err)
 		case h := <-c:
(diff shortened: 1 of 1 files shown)


sae/rpc.go
```diff
@@ -63,14 +64,24 @@ func (b *ethAPIBackend) GetTd(context.Context, common.Hash) *big.Int {
...
 
+func (b *ethAPIBackend) HeaderByNumber(ctx context.Context, n rpc.BlockNumber) (*types.Header, error) {
+	return readByNumber(b, n, rawdb.ReadHeader)
+}
+
 func (b *ethAPIBackend) BlockByNumber(ctx context.Context, n rpc.BlockNumber) (*types.Block, error) {
+	return readByNumber(b, n, rawdb.ReadBlock)
+}
+
+type canonicalReader[T any] func(ethdb.Reader, common.Hash, uint64) *T
```
ARR4N: Nice

ARR4N (changes requested):

PR #63, opened 2026-01-02
test: fuzz integration of `worstcase` with VM consensus+execution

ARR4N:
Not an actual `test Fuzz` as it is long-lived, but a property-based test to try and trigger a failure of the `worstcase` package once the block is actually executed (i.e. get `core.ApplyTransaction()` to return an error, resulting in a `FATAL` log by the `saexec.Executor`).

Additionally, block-building now propagates the worst-case gas price and sender balances to be checked immediately prior to execution. Violation of the bounds is logged at `ERROR` but optimistically ignored as it might just be a near miss. However, the use of `saetest.TBLogger` means that any such log will cause a test failure even if not a production one.

Local run of approximately 80 million transactions:
```bash
$ git rev-parse HEAD && go test -run WorstCase ./sae -test.timeout 60m -worstcase.fuzz.blocks 100_000 -worstcase.fuzz.parallel 16
be65077c682ebee2c7021497c24c1d3c2af4920d
ok      github.com/ava-labs/strevm/sae  1323.320s
```

### DIFF
--- a/worstcase/state.go
+++ b/worstcase/state.go
@@ -131,7 +131,8 @@ func (s *State) StartBlock(h *types.Header) error {
 	}
 
 	s.baseFee = s.clock.BaseFee()
-	s.minOpBurnerBalances = s.minOpBurnerBalances[:0] // [State.FinishBlock] returns a clone so we can reuse the alloc here
+	clear(s.minOpBurnerBalances) // [State.FinishBlock] returns a clone so we can reuse the alloc here
+	s.minOpBurnerBalances = s.minOpBurnerBalances[:0]
 
 	// expectedParentHash is updated prior to modifying the GasLimit and BaseFee
 	// to ensure that historical block hashes are not modified.


blocks/invariants.go
```diff
@@ -6,10 +6,86 @@ package blocks
...
+		b.log.Error("Actual base fee > predicted worst case",
+			zap.Stringer("actual", actual),
+			zap.Stringer("predicted", b.bounds.MaxBaseFee),
+		)
+
+	case 0: // Coverage visualisation
+		_ = 0
+	case -1:
+		_ = 0
+	}
```
StephenButtolph: Why do we use `Cmp` rather than the more natural methods?
```suggestion
	if actual.Gt(b.bounds.MaxBaseFee) {
		b.log.Error("Actual base fee > predicted worst case",
			zap.Stringer("actual", actual),
			zap.Stringer("predicted", b.bounds.MaxBaseFee),
		)
	}
```

sae/blocks.go
```diff
@@ -186,14 +187,23 @@ func (vm *VM) buildBlock(
...
+		}
+		logTo("Could not start worst-case block calculation",
 			zap.Error(err),
 		)
-		return nil, fmt.Errorf("starting worst-case state for new block: %v", err)
+		return nil, fmt.Errorf("starting worst-case state for new block: %w", err)
 	}
 
+	bounds := &blocks.WorstCaseBounds{
+		MaxBaseFee: new(uint256.Int).Set(state.BaseFee()),
```
StephenButtolph: This is a bit of a general comment throughout the SAE codebase. We copy stuff _really_ aggressively (Sometimes copying immediately after copying).

I think the codebase is small enough that it would make the most sense to just do a single comprehensive pass of minor performance cleanups though.

sae/blocks.go
```diff
@@ -186,14 +187,23 @@ func (vm *VM) buildBlock(
...
+			logTo = log.Debug
+		}
+		logTo("Could not start worst-case block calculation",
 			zap.Error(err),
 		)
-		return nil, fmt.Errorf("starting worst-case state for new block: %v", err)
+		return nil, fmt.Errorf("starting worst-case state for new block: %w", err)
 	}
 
+	bounds := &blocks.WorstCaseBounds{
```
StephenButtolph: I feel like `WorstCaseBounds` should be a result from `worstcase.State`, rather than being built during block building.

Perhaps `worstcase.State.FinishBlock` should return this?

sae/worstcase_test.go
```diff
@@ -0,0 +1,237 @@
...
+			seed := rand.Uint64() //nolint:gosec // Not for security
+			t.Logf("RNG seed: %d", seed)
+			rng := rand.New(rand.NewPCG(0, seed)) //nolint:gosec // Allow for reproducibility
+
+			for range flags.numBlocks {
+				for range rng.UintN(flags.maxNewTxsPerBlock) {
+					from := rng.IntN(numEOAs)
+					to := rng.IntN(numEOAs + 1)
+					// TODO(arr4n) why does increasing this slow everything
+					// down? Without parallel tests:
```
StephenButtolph: > this

meaning:
```suggestion
					// TODO(arr4n) why does increasing maxGasLimit slow
					// everything down? Without parallel tests:
```
?

blocks/invariants.go
ARR4N: My (subjective) assessment is that these bounds are some of the most critical things to test in the entire implementation, so I wanted greater visibility into how they're behaving in integration tests. The `Cmp` is to allow the coverage visualisation in the other `case`s (i.e. to know that we do sometimes get it exactly correct, which is particularly informative w.r.t. balances).

sae/vm_test.go
```diff
@@ -84,39 +92,49 @@ func newSUT(tb testing.TB, numAccounts uint, opts ...sutOption) (context.Context
...
+		logLevel: logging.Warn,
+		alloc:    saetest.MaxAllocFor(wallet.Addresses()...),
 		genesisOptions: []blockstest.GenesisOption{
 			blockstest.WithTimestamp(saeparams.TauSeconds),
 		},
 	}, opts...)
 
+	// If the options updated the ChainConfig it could require a new signer.
+	// This is unavoidable because we need the wallet earlier, to determine the
+	// addresses for the alloc.
```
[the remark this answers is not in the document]
ARR4N: #68

blocks/invariants.go
StephenButtolph: Personally, the coverage is useless to me (because people only ever look at it when initially writing it, it isn't actually part of CI).

blocks/block.go
```diff
@@ -37,6 +37,9 @@ type Block struct {
 	// Only the genesis block or the last pre-SAE block is synchronous. These
 	// are self-settling by definition so their `ancestry` MUST be nil.
 	synchronous bool
+	// Determined during block building and MUST be set before execution as
+	// expected by the Executor.
+	bounds *WorstCaseBounds
```
StephenButtolph: I would have sworn that I commented this on the first pass.

Why do we require this to be set? I feel like it would be cleaner to just skip the sanity checks if the bounds aren't set.

Then we could just remove `SetUninformativeWorstCaseBounds`.

blocks/invariants.go
ARR4N: I'll revisit it later. Tracking in #71.

sae/blocks.go
```diff
@@ -251,7 +259,12 @@ func (vm *VM) buildBlock(
 		included,
 		receipts,
 	)
-	return vm.newBlock(ethB, parent, lastSettled)
+	b, err := vm.newBlock(ethB, parent, lastSettled)
+	if err != nil {
+		return nil, err
+	}
+	b.SetWorstCaseBounds(bounds)
```
StephenButtolph: I don't know that I actually want you to change this, but this does feel like `newBlock` could take in the bounds

PR #64, opened 2026-01-02
feat: `FilterAPI` for WebSocket subscription support

StephenButtolph:
This implements all the `Subscribe*` functions on the API backend and instantiates the filter API.

Some of the subscribe functions don't make sense for SAE, so they are noops. Unlike in the PoC, the noop subscriptions are _not_ closed when the VM is closed. This aligns with the behavior of the rest of the subscriptions (and the normal geth behavior).

### DIFF
--- a/sae/vm_test.go
+++ b/sae/vm_test.go
@@ -523,7 +523,10 @@ func TestSubscriptions(t *testing.T) {
 	newHeads := make(chan *types.Header, 1)
 	sub, err := sut.SubscribeNewHead(ctx, newHeads)
 	require.NoError(t, err, "SubscribeNewHead(...)")
-	t.Cleanup(sub.Unsubscribe)
+	// The subscription is closed in a defer rather than via t.Cleanup to ensure
+	// that is is closed before the rest of the SUT is torn down. Otherwise,
+	// there could be a goroutine leak.
+	defer sub.Unsubscribe()
 
 	b := sut.runConsensusLoop(t, sut.lastAcceptedBlock(t))
 	got := <-newHeads


sae/rpc.go
```diff
@@ -101,3 +107,56 @@ func (b *ethAPIBackend) resolveBlockNumber(bn rpc.BlockNumber) (uint64, error) {
...
+}
+
+func (b *ethAPIBackend) SubscribeLogsEvent(ch chan<- []*types.Log) event.Subscription {
+	return b.vm.exec.SubscribeLogsEvent(ch)
+}
+
+func (b *ethAPIBackend) SubscribePendingLogsEvent(chan<- []*types.Log) event.Subscription {
+	// In SAE, "pending" refers to the execution status. There are no logs known
+	// for transactions pending execution.
+	return newNoopSubscription()
```
StephenButtolph: I spent awhile looking into whether or not this should be treated the same as `SubscribeLogsEvent`.

In coreth and subnet-evm, we do not support pending logs events.

Also, in geth this is actually not even a documented API: https://geth.ethereum.org/docs/interacting-with-geth/rpc/pubsub#logs

I suspect it is _highly_ unexpected for a user to listen exclusively to pending logs events. So returning the normal logs events here will just result in duplicated notifications.

ARR4N: SGTM

sae/vm_test.go
```diff
@@ -121,9 +123,9 @@ func newSUT(tb testing.TB, numAccounts uint, opts ...sutOption) (context.Context
 
 	handlers, err := snow.CreateHandlers(ctx)
 	require.NoErrorf(tb, err, "%T.CreateHandlers()", snow)
-	server := httptest.NewServer(handlers[rpcHTTPExtensionPath])
+	server := httptest.NewServer(handlers[wsHTTPExtensionPath])
 	tb.Cleanup(server.Close)
-	client, err := ethclient.Dial(server.URL)
+	client, err := ethclient.Dial("ws://" + server.Listener.Addr().String())
```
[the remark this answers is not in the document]
ARR4N: > But I think the ws client is a superset of the http client

I think so too. My PoC tests also only used ws, based on similar logic.

ARR4N (approved):

PR #69, opened 2026-01-04
fix: `WaitUntil{Executed,Settled}()` wait for atomic pointers

ARR4N:
Although there was no guarantee that `blocks.Block.WaitExecuted()` returning `nil` error would mean that `saexec.Executor.LastExecuted()` reflected this change in state (both being internal indicators), it was a reasonable assumption and led to flakes fixed in #62.This PR now provides such a guarantee and documents the "atomics before channel closes" heuristic.

The rationale for the heuristic is that it's reasonable for a consumer receiving a broadcast to assume that a subsequent poll for related information reflects what they received.


StephenButtolph (approved):
lgtm with or without the varargs of atomics.

PR #73, opened 2026-01-10
feat: `LastToSettleAt()` supports gas-resolution time

ARR4N:
Everything in this PR is motivated by a change to `blocks.LastToSettleAt()` to accept a `proxytime.Time[gas.Gas]` instead of a Unix timestamp. Without this change, settlement could be up to (but excluding) 6s in the past if a new block has a fractional time near 1.

Starting at the changes to `LastToSettleAt()` provides full context for all other changes.

### DIFF
--- a/blocks/settlement_test.go
+++ b/blocks/settlement_test.go
@@ -29,7 +29,7 @@ import (
 //nolint:testableexamples // Output is meaningless
 func ExampleRange() {
 	parent := blockBuildingPreference()
-	settle, ok, err := LastToSettleAt(vmHooks(), time.Now().Add(-params.TauSeconds), parent)
+	settle, ok, err := LastToSettleAt(vmHooks(), time.Now().Add(-params.Tau), parent)
 	if err != nil {
 		// Due to a malformed input to block verification.
 		return // err


blocks/settlement.go
```diff
@@ -234,28 +239,45 @@ func LastToSettleAt(settleAt uint64, parent *Block) (b *Block, ok bool, _ error)
...
 		}
-		// TODO(arr4n) more fine-grained checks are possible by computing the
-		// minimum possible gas consumption of blocks. For example,
-		// `block.BuildTime()+block.intrinsicGasSum()` can be compared against
-		// `settleAt`, as can the sum of a chain of blocks.
+		{
+			e := &block.interimExecution
+			e.RLock()
+			passed := e.Time != nil && e.Compare(settleAt) > 0
+			e.RUnlock()
```
StephenButtolph: Super nit, but I feel like this would be a nice helper function. I'm always a fan of setting up the locking so that we can throw an unlock into a defer.

sae/blocks.go
```diff
@@ -106,23 +109,31 @@ func (vm *VM) buildBlock(
...
 	}
 
+	// Even though we need a [proxytime.Time] value, we only have a gas target
+	// and not a rate. We therefore use [gastime] for construction as it is the
+	// source of truth for target-to-rate conversion.
+	bTime := gastime.OfBlock(vm.hooks, hdr, parent.Header(), 0).Time
+	// If `parent` is the genesis block then the notion of a grandparent is
+	// undefined, so we have to construct its time ourselves.
+	pTime := proxytime.New(parent.BuildTime(), bTime.Rate())
+	pTime.Tick(vm.hooks.SubSecondBlockTime(bTime.Rate(), parent.Header()))
```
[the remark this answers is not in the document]
ARR4N: `SubSecondBlockTime()` now returns a `time.Duration` so these issues disappear.

gastime/gastime.go
```diff
@@ -51,8 +54,41 @@ func (tm *Time) establishInvariants() {
...
+// provided by the hooks.
+func OfBlock(hooks hook.Points, hdr, parent *types.Header, startingExcess gas.Gas) *Time {
+	target := hooks.GasTargetAfter(parent)
+	return newT(
+		hdr.Time,
+		subSecondGasDuration(hooks, hdr, SafeRateOfTarget(target)),
+		target,
+		startingExcess,
+	)
+}
```
StephenButtolph: This feels weird to me. We never use this for a `gastime.Time`, we always use this just for a `proxytime.Time`.

Can we replace this just with constructing the proxytime?

proxytime/proxytime_test.go
```diff
@@ -143,51 +143,103 @@ func TestSetRate(t *testing.T) {
...
-			wantTruncated: frac(0, 1),
+			wantRoundedUp: frac(0, 1),
 			wantInvariant: 3_000,
 		},
 		{
-			newRate:       75,
-			wantNumerator: 7,                   // 7.5
-			wantTruncated: frac(7_500, 15_000), // rounded down by 0.5, denominated in the old rate
+			newRate:       75,                // 15_000 / 20
+			wantNumerator: 8,                 // ceil(1_500/20 == 7.5)
```
StephenButtolph:
```suggestion
			newRate:       75,                // 15_000 / 200
			wantNumerator: 8,                 // ceil(1_500/200 == 7.5)
```

gastime/gastime.go
ARR4N: I agree and have always felt that too. It's now `blocks.GasTime() *proxytime.Time[gas.Gas]`.

Originally I thought that placing it in `gastime` was required in order to have access to `SafeRateOfTarget()` without introducing a circular dependency with `hook`, but I hadn't thought of placing it in the `blocks` package, which imports both of those packages! 🤦

gastime/gastime.go
```diff
@@ -51,8 +54,41 @@ func (tm *Time) establishInvariants() {
 // `target` * [TargetToRate] units of [gas.Gas] is equivalent to a tick of 1
 // second. Targets are clamped to [MaxTarget].
 func New(unixSeconds uint64, target, startingExcess gas.Gas) *Time {
+	return newT(unixSeconds, 0, target, startingExcess)
+}
```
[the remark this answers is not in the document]
ARR4N: Everything introduced in this PR has been removed with the refactor to `blocks.GasTime()`. However `makeTime()` remains to ensure that `New()` and `Clone()` set the invariants in the same way.

proxytime/proxytime.go
```diff
@@ -169,31 +194,45 @@ func (tm *Time[D]) SetRate(hertz D) (truncated FractionalSecond[D], err error) {
...
+// specified one. See [Time.SetRate] for details about overflow errors and
+// rounding.
+func (tm *Time[D]) scale(val, newRate D) (scaled D, roundedUp FractionalSecond[D], err error) {
+	scaled, round, err := intmath.MulDivCeil(val, newRate, tm.hertz)
 	if err != nil {
 		return 0, FractionalSecond[D]{}, fmt.Errorf("scaling %d from rate of %d to %d: %w", val, tm.hertz, newRate, err)
 	}
-	return scaled, FractionalSecond[D]{Numerator: trunc, Denominator: tm.hertz}, nil
+	return scaled, FractionalSecond[D]{Numerator: round / newRate, Denominator: tm.hertz}, nil
+}
```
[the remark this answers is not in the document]
ARR4N: Removal makes sense. I shouldn't have ever introduced it as it was, as you point out, YAGNI.

FWIW, there were a couple of reasons I chose to return at the original rate:

1. The denominator in your scenario would have to be the LCM of the new and old rates. This could very easily overflow `uint64` with just two prime numbers each ~1e10.
2. The margin of error in the original rate is greater than the error in the returned `truncated` / `roundedUp`. In your concrete example, $\frac{3}{600} < \frac{0.5}{25}$.

proxytime/proxytime_test.go
```diff
@@ -143,51 +143,103 @@ func TestSetRate(t *testing.T) {
...
+			wantRoundedUp: frac(100, 15_000), // 8/75 == 1_600/15_000 -> 1600-1500 rounding up
 			wantInvariant: 15,
 		},
+		{
+			newRate:       25,          // 75 / 3
+			wantNumerator: 3,           // ceil(8/3 == 2.66...)
+			wantRoundedUp: frac(1, 75), // 3/25 == 9/75 -> 9-8 rounding up
+			wantInvariant: 5,
+		},
```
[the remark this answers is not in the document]
ARR4N: See point (2) of rationale in [my response](https://github.com/ava-labs/strevm/pull/73#discussion_r2718710060). It's entirely academic now, though, as I'm removing it.

proxytime/proxytime.go
```diff
@@ -162,38 +186,52 @@ func (tm *Time[D]) SetRate(hertz D) (truncated FractionalSecond[D], err error) {
...
 }
 
 // scale returns `val`, scaled from the existing [Time.Rate] to the newly
-// specified one. See [Time.SetRate] for details about truncation and overflow
-// errors.
-func (tm *Time[D]) scale(val, newRate D) (scaled D, truncated FractionalSecond[D], err error) {
-	scaled, trunc, err := intmath.MulDiv(val, newRate, tm.hertz)
+// specified one. See [Time.SetRate] for details about overflow errors and
+// rounding.
+func (tm *Time[D]) scale(val, newRate D) (scaled D, roundedUp FractionalSecond[D], err error) {
```
StephenButtolph: I think we can remove `roundedUp` from here as well

PR #76, opened 2026-01-13
refactor: `newNetworkedSUTs()`

ARR4N:
The `newNetworkedSUTs()` function was a bit unwieldy so I abstracted it into smaller chunks. There's no new functionality (at least not intentionally).

I haven't documented all of the new methods just yet as I wanted to get @StephenButtolph's first impressions before doing so.


StephenButtolph (commented):
seems reasonable to me, definitely nice to move some of the complexity to methods

Issue #77, opened 2026-01-16
Make ACP-176 math configurable


ceyonur:
(Moved From https://github.com/ava-labs/avalanchego/issues/4285)

The existing acp-176 calculations in `gastime` [package](https://github.com/ava-labs/strevm/tree/main/gastime/gastime.go) only supports the existing C-Chain configuration. ACP-224 involves incorporating ACP-176 into Subnet-EVM chains, but allowing them to set their own fee configurations. 

Add a ACP-176 config/params that allows for the following variables (currently hard coded as constants) to be configured by various chains:
```
type Config struct {
	MinGasPrice        gas.Price // M
	TimeToDouble       uint64    // in seconds
}
```

The `TargetToExcessScaling` and the constant `1` wei in  `(tm *Time) Price()` will then need to be made functions based on a given configuration. 

`TargetToExcessScaling` (`K`) should become calculatable from `TimeToDouble` as in specified in [ACP-176](https://github.com/avalanche-foundation/ACPs/blob/main/ACPs/176-dynamic-evm-gas-limit-and-price-discovery-updates/README.md#making-t-dynamic). `K` was chosen 87 such that at sustained maximum capacity ( T ∗ 2 gas/second), the fee rate will double every ~60.3 seconds. (87 ~= 60 / ln(2)). Making it `K=~TimeToDouble/ln(2)`

For `K/TimeToDouble` calculations, a rational approximation of ln(2) can be used to ensure precision and avoiding using floating point math.

For more information see ACP-224 [here](https://github.com/avalanche-foundation/ACPs/tree/main/ACPs/224-dynamic-gas-limit-in-subnet-evm#acp-176-parameters-in-subnet-evm).

PR #80, opened 2026-01-22
feat: Make minPrice and excessScalingFactor Configurable

ceyonur:
This PR makes minPrice and excessScalingFactor Configurable, also adds a static pricing mode. These parameters will be required by ACP-224, which intends to set these via a system contract. 

Closes: #77

### DIFF
--- a/sae/block_builder.go
+++ b/sae/block_builder.go
@@ -193,19 +193,19 @@ func (b *blockBuilderG[T]) buildWithTxs(
 
 	unsettled := blocks.Range(lastSettled, parent)
 	for _, block := range unsettled {
-		log := log.With(
+		blockLog := log.With(
 			zap.Uint64("block_height", block.Height()),
 			zap.Stringer("block_hash", block.Hash()),
 		)
 		if err := state.StartBlock(block.Header()); err != nil {
-			log.Warn("Could not start historical worst-case calculation",
+			blockLog.Warn("Could not start historical worst-case calculation",
 				zap.Error(err),
 			)
 			return nil, fmt.Errorf("starting worst-case state for block %d: %v", block.Height(), err)
 		}
 		for i, tx := range block.Transactions() {
 			if err := state.ApplyTx(tx); err != nil {
-				log.Warn("Could not apply tx during historical worst-case calculation",
+				blockLog.Warn("Could not apply tx during historical worst-case calculation",
 					zap.Int("tx_index", i),
 					zap.Stringer("tx_hash", tx.Hash()),
 					zap.Error(err),
@@ -215,14 +215,14 @@ func (b *blockBuilderG[T]) buildWithTxs(
 		}
 		ops, err := b.hooks.EndOfBlockOps(block.EthBlock())
 		if err != nil {
-			log.Warn("Could not extract ops during historical worst-case calculation",
+			blockLog.Warn("Could not extract ops during historical worst-case calculation",
 				zap.Error(err),
 			)
 			return nil, fmt.Errorf("extracting ops of block %d to worst-case state: %v", block.Height(), err)
 		}
 		for i, op := range ops {
 			if err := state.Apply(op); err != nil {
-				log.Warn("Could not apply op during historical worst-case calculation",
+				blockLog.Warn("Could not apply op during historical worst-case calculation",
 					zap.Int("op_index", i),
 					zap.Stringer("op_id", op.ID),
 					zap.Error(err),
@@ -231,7 +231,7 @@ func (b *blockBuilderG[T]) buildWithTxs(
 			}
 		}
 		if _, err := state.FinishBlock(); err != nil {
-			log.Warn("Could not finish historical worst-case calculation",
+			blockLog.Warn("Could not finish historical worst-case calculation",
 				zap.Error(err),
 			)
 			return nil, fmt.Errorf("finishing worst-case state for block %d: %v", block.Height(), err)
--- a/saexec/execution.go
+++ b/saexec/execution.go
@@ -51,7 +51,7 @@ func (e *Executor) Enqueue(ctx context.Context, block *blocks.Block) error {
 	}
 }
 
-var emergencyPlaybookLink = zap.String("playbook", "https://github.com/ava-labs/strevm/issues/28")
+const emergencyPlaybookLink = "https://github.com/ava-labs/strevm/issues/28"
 
 func (e *Executor) processQueue() {
 	defer close(e.done)
@@ -62,27 +62,39 @@ func (e *Executor) processQueue() {
 			return
 
 		case block := <-e.queue:
-			logger := e.log.With(
+			log := e.log.With(
 				zap.Uint64("block_height", block.Height()),
 				zap.Uint64("block_time", block.BuildTime()),
 				zap.Stringer("block_hash", block.Hash()),
 				zap.Int("tx_count", len(block.Transactions())),
 			)
 
-			if err := e.execute(block, logger); err != nil {
-				logger.Fatal(
-					"Block execution failed; see emergency playbook",
-					emergencyPlaybookLink,
+			err := e.execute(block, log)
+			switch {
+			case errors.Is(err, errFatal):
+				log.Fatal( //nolint:gocritic // False positive, will not terminate the process
+					"Block execution failed",
+					zap.String("playbook", emergencyPlaybookLink),
 					zap.Error(err),
 				)
+			case err != nil:
+				log.Error(
+					"Error of unknown severity in block execution",
+					zap.String("if_escalation_required", emergencyPlaybookLink),
+					zap.Error(err),
+				)
+			}
+			if err != nil {
 				return
 			}
 		}
 	}
 }
 
-func (e *Executor) execute(b *blocks.Block, logger logging.Logger) error {
-	logger.Debug("Executing block")
+var errFatal = errors.New("fatal execution error")
+
+func (e *Executor) execute(b *blocks.Block, log logging.Logger) error {
+	log.Debug("Executing block")
 
 	// Since `b` hasn't been executed, it definitely hasn't been settled, so we
 	// are guaranteed to have a non-nil parent available.
@@ -122,11 +134,6 @@ func (e *Executor) execute(b *blocks.Block, logger logging.Logger) error {
 		stateDB.SetTxContext(tx.Hash(), ti)
 		b.CheckSenderBalanceBound(stateDB, signer, tx)
 
-		logger = logger.With(
-			zap.Int("tx_index", ti),
-			zap.Stringer("tx_hash", tx.Hash()),
-		)
-
 		receipt, err := core.ApplyTransaction(
 			e.chainConfig,
 			e.chainContext,
@@ -139,12 +146,7 @@ func (e *Executor) execute(b *blocks.Block, logger logging.Logger) error {
 			vm.Config{},
 		)
 		if err != nil {
-			logger.Fatal(
-				"Transaction execution errored (not reverted); see emergency playbook",
-				emergencyPlaybookLink,
-				zap.Error(err),
-			)
-			return err
+			return fmt.Errorf("%w: transaction execution errored (not reverted) [%d](%#x): %v", errFatal, ti, tx.Hash(), err)
 		}
 
 		perTxClock.Tick(gas.Gas(receipt.GasUsed))
@@ -180,12 +182,7 @@ func (e *Executor) execute(b *blocks.Block, logger logging.Logger) error {
 	numTxs := len(b.Transactions())
 	ops, err := e.hooks.EndOfBlockOps(b.EthBlock())
 	if err != nil {
-		logger.Fatal(
-			"Unable to extract extra block operations; see emergency playbook",
-			emergencyPlaybookLink,
-			zap.Error(err),
-		)
-		return err
+		return fmt.Errorf("%w: %T.EndOfBlockOps(%#x): %v", errFatal, e.hooks, b.Hash(), err)
 	}
 	for i, o := range ops {
 		b.CheckOpBurnerBalanceBounds(stateDB, numTxs+i, o)
@@ -194,14 +191,7 @@ func (e *Executor) execute(b *blocks.Block, logger logging.Logger) error {
 		b.SetInterimExecutionTime(perTxClock)
 
 		if err := o.ApplyTo(stateDB); err != nil {
-			logger.Fatal(
-				"Extra block operation errored; see emergency playbook",
-				zap.Int("op_index", i),
-				zap.Stringer("op_id", o.ID),
-				emergencyPlaybookLink,
-				zap.Error(err),
-			)
-			return err
+			return fmt.Errorf("%w: applying end-of-block operation [%d](%v): %v", errFatal, i, o.ID, err)
 		}
 	}
 
@@ -211,7 +201,7 @@ func (e *Executor) execute(b *blocks.Block, logger logging.Logger) error {
 		return fmt.Errorf("after-block gas time update: %w", err)
 	}
 
-	logger.Debug(
+	log.Debug(
 		"Block execution complete",
 		zap.Uint64("gas_consumed", uint64(blockGasConsumed)),
 		zap.Time("gas_time", gasClock.AsTime()),


gastime/gastime.go
```diff
@@ -90,19 +127,47 @@ func (tm *Time) Excess() gas.Gas {
...
+	return gas.CalculatePrice(tm.minPrice, tm.excess, tm.excessScalingFactor())
 }
 
-// excessScalingFactor returns the K variable of ACP-103/176, i.e. 87*T, capped
-// at [math.MaxUint64].
+// excessScalingFactor returns the K variable of ACP-103/176, i.e.
+// targetToExcessScaling*T, capped at [math.MaxUint64].
 func (tm *Time) excessScalingFactor() gas.Gas {
-	const overflowThreshold = math.MaxUint64 / TargetToExcessScaling
+	// TODO (cey): Should we verify this is non-zero instead?
```
ceyonur: todo here. We can resolve it in this PR.

gastime/gastime.go
```diff
@@ -119,19 +152,133 @@ func (tm *Time) Excess() gas.Gas {
...
+//
+// When [config.TargetToExcessScaling] is MaxUint64, this is "fixed price mode" and
+// always returns minPrice regardless of excess. This prevents the fake
+// exponential approximation from producing e^(x/K) ≈ e^1 when excess is also
+// very large.
 func (tm *Time) Price() gas.Price {
-	return gas.CalculatePrice(1 /* M */, tm.excess, tm.excessScalingFactor())
+	// We need to check explicitly this edge case since excess can be very large and the fake
+	// exponential approximation can produce e^(x/K) ≈ e^1, which would cause the price to change.
+	if tm.config.targetToExcessScaling == math.MaxUint64 {
```
ceyonur: This is a weird check here, but otherwise we actually let price to be changed with a high excess.

gastime/testdata/fuzz/FuzzPriceInvarianceAfterBlock/036482440c1ec875
alarso16: Is it conventional to commit fuzz data?

gastime/acp176.go
```diff
@@ -21,13 +21,23 @@ func (tm *Time) BeforeBlock(hooks hook.Points, h *types.Header) {
...
-// sourced from [hook.Points] and [types.Header].
+// AfterBlock is intended to be called after processing a block, with the
+// gas configuration sourced from [hook.Points] and [types.Header].
 func (tm *Time) AfterBlock(used gas.Gas, hooks hook.Points, h *types.Header) error {
 	tm.Tick(used)
-	target := hooks.GasTargetAfter(h)
-	if err := tm.SetTarget(target); err != nil {
+	cfg := hooks.GasConfigAfter(h)
+
+	// Set target first as we use the target to calculate the excess scaling factor.
```
alarso16: You say first, in reference to what? Setting the options?

gastime/gastime.go
```diff
@@ -33,13 +35,31 @@ type Time struct {
 	TimeMarshaler
 }
 
+// An Option configures the [Time] created by [New].
+type Option = options.Option[config]
+
+// WithTargetToExcessScaling overrides the target to excess scaling ratio.
```
alarso16: Will you add here that `math.MaxUint64` is a special value? Seems like the best place to document it

gastime/gastime.go
```diff
@@ -119,19 +152,133 @@ func (tm *Time) Excess() gas.Gas {
...
+
+// findExcessForPrice uses binary search over uint64 to find an excess value
+// that produces targetPrice with the given minPrice and excessScalingFactor (K).
+//
+// The price formula is: P = M * e^(x / K), where K = targetToExcessScaling * target.
+//   - P is the price (targetPrice)
+//   - M is the minimum price (minPrice)
+//   - x is the excess
+//   - targetToExcessScaling is the target to excess scaling ratio
+//   - target is the target gas consumption per second
```
alarso16: It seems reasonable to put the actual formula for readers to follow along:

```suggestion
//   - target is the target gas consumption per second
// Thus we are solving for: x = K * ln(P / m)
```

gastime/gastime.go
```diff
@@ -119,19 +152,133 @@ func (tm *Time) Excess() gas.Gas {
...
+func findExcessForPrice(targetPrice, minPrice gas.Price, targetToExcessScaling gas.Gas, target gas.Gas) gas.Gas {
+	// Check edge cases:
+	// If the target price is less than the minimum price, or the target to excess scaling ratio is 0 or MaxUint64,
+	// return 0.
+	// Even though we return 0 for excess it won't avoid accumulating excess in the long run.
+	if targetPrice <= minPrice ||
+		targetToExcessScaling == math.MaxUint64 ||
+		targetToExcessScaling == 0 {
```
alarso16: Wasn't this verified to not be true already?

gastime/acp176_test.go
```diff
@@ -13,9 +13,71 @@ import (
...
+			require.Error(t, err)
+
+			// Config unchanged after rejected update
+			assert.Equal(t, initialScaling, tm.config.targetToExcessScaling)
+			assert.Equal(t, initialMinPrice, tm.config.minPrice)
+		})
+	}
+}
+
+func ptr[T any](v T) *T { return &v }
```
alarso16: You should use the avalanchego one. Import "github.com/ava-labs/avalanchego/utils", it's called `utils.PointerTo`

gastime/acp176.go
```diff
@@ -21,13 +21,23 @@ func (tm *Time) BeforeBlock(hooks hook.Points, h *types.Header) {
...
+	if err := tm.SetTarget(cfg.Target); err != nil {
 		return fmt.Errorf("%T.SetTarget() after block: %w", tm, err)
 	}
+	// TargetToExcessScaling and MinPrice are optional, so we apply them as options.
+	opts := GasConfigToOpts(cfg)
+	if len(opts) > 0 {
+		if err := tm.SetOpts(opts...); err != nil {
+			return err
+		}
+	}
```
ARR4N: As with my comment on `MarkSynchronous()`, this suggests that the type's API signatures are difficult to use and need to be refactored.

gastime/gastime.go
```diff
@@ -119,19 +153,133 @@ func (tm *Time) Excess() gas.Gas {
...
+	}
+
+	// Calculate new K with new scaling and current target
+	k := excessScalingFactorOf(targetToExcessScaling, target)
+
+	// Binary search over [0, maxExcessSearchCap(k)] for smallest excess where price >= targetPrice.
+	maxTargetExcess := maxExcessSearchCap(k)
+	lo, hi := uint64(0), uint64(maxTargetExcess)
+	for lo < hi {
+		mid := lo + (hi-lo)>>1 // avoid overflow
```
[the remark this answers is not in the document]
ceyonur: This was a kind-of self-comment for myself as I managed to overflow this lol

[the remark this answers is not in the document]
StephenButtolph: oldie but a goodie https://medium.com/@khaledalokby/binary-search-bug-did-you-know-it-existed-in-java-3ba2aa06e3dc

gastime/gastime.go
```diff
@@ -119,19 +121,120 @@ func (tm *Time) Excess() gas.Gas {
 	return tm.excess
 }
 
+var (
+	errTargetToExcessScalingZero = errors.New("targetToExcessScaling must be non-zero")
+	errMinPriceZero              = errors.New("minPrice must be non-zero")
+)
+
+func (c *config) validate() error {
+	if c.targetToExcessScaling == 0 {
```
alarso16: Can we validate here the `maxExcessSearchCap` won't overflow? Seems like it's more straightforward, is there an edge case where this would act strangely?

gastime/gastime.go
```diff
@@ -119,19 +121,120 @@ func (tm *Time) Excess() gas.Gas {
...
+	if targetPrice <= tm.config.minPrice || tm.config.staticPricing {
+		return 0
+	}
+
+	k := tm.excessScalingFactor()
+
+	// Binary search over [0, maxExcessSearchCap(k)] for smallest excess where price >= targetPrice.
+	lo, hi := gas.Gas(0), maxExcessSearchCap(k)
+	for lo < hi {
+		mid := lo + (hi-lo)>>1
```
alarso16: Why not divide by two? Is the performance difference significant enough?

hook/hookstest/stub.go
```diff
@@ -21,13 +22,51 @@ import (
...
+// It uses [hook.DefaultGasConfig] unless overridden by [WithGasConfig].
+func NewStub(target gas.Gas, opts ...HookOption) *Stub {
+	s := &Stub{
+		Target:    target,
+		GasConfig: hook.DefaultGasConfig(),
+	}
+	for _, opt := range opts {
+		options.ApplyTo(s, opt)
+	}
+	return s
```
StephenButtolph: It feels weird not to use the varargs for `options.ApplyTo`
```suggestion
	return options.ApplyTo(&Stub{
		Target:    target,
		GasConfig: hook.DefaultGasConfig(),
	}, opts...)
```

hook/hook.go
```diff
@@ -135,6 +138,35 @@ func (o *Op) ApplyTo(stateDB *state.StateDB) error {
...
 }
 
+// GasConfig contains gas-related parameters that can be configured via hooks.
+type GasConfig struct {
+	// TargetToExcessScaling is the ratio between the gas target and the
+	// reciprocal of the excess coefficient used in price calculation
+	// (K variable in ACP-176, where K = TargetToExcessScaling * T).
+	TargetToExcessScaling gas.Gas
+	// MinPrice is the minimum gas price / base fee (M parameter in ACP-176).
+	MinPrice gas.Price
```
StephenButtolph: Should we document here that 0 is invalid for both of these?

Also, should we consider allowing `MinPrice == 0` (maybe only when `StaticPricing == true`)? I hate when people ask for "how do we disable fees on our L1", but I don't think it would be too much work for us to support them in this case.

gastime/gastime.go
```diff
@@ -119,19 +121,120 @@ func (tm *Time) Excess() gas.Gas {
...
+	// by setting the excess to 0. ( P = M * e^(0 / K) = M )
+	// Note: Even though we return 0 for excess it won't avoid accumulating excess in the long run.
+	if targetPrice <= tm.config.minPrice || tm.config.staticPricing {
+		return 0
+	}
+
+	k := tm.excessScalingFactor()
+
+	// Binary search over [0, maxExcessSearchCap(k)] for smallest excess where price >= targetPrice.
+	lo, hi := gas.Gas(0), maxExcessSearchCap(k)
```
StephenButtolph: Similar comment w.r.t. if this is an optimization or not. I believe it is an optimization (or perhaps inspired from avalanchego's `maxTargetExcess` usage)... Should we be benchmarking this function?

In avalanchego I mainly calculated the cap so that I could use `sort.Search` (which works over `int` rather than `uint64`)... But here we could feasibly hit `uint64` values, so we need to implement our own binary search anyways.

gastime/gastime.go
```diff
@@ -119,19 +121,120 @@ func (tm *Time) Excess() gas.Gas {
...
+// Because x = ln(P/M) * K, we know x <= ln(MaxUint64/M) * K <= 45 * K
+// However, if 45 * K would overflow, MaxUint64 is returned
+func maxExcessSearchCap(k gas.Gas) gas.Gas {
+	const maxExcessMultiplier = 45 // ≈ ceil(ln(MaxUint64))
+	if k > math.MaxUint64/gas.Gas(maxExcessMultiplier) {
 		return math.MaxUint64
 	}
-	return TargetToExcessScaling * tm.target
+	return gas.Gas(maxExcessMultiplier) * k
+}
```
StephenButtolph: Should we add a `BoundedMultiply` to the intmath package? `excessScalingFactorOf` and `maxExcessSearchCap` are both just `min(a*b, MaxUint64)` without overflow

gastime/gastime.go
StephenButtolph: > Is the performance difference significant enough?

There is no performance difference, the compiler produces the same assembly code

gastime/gastime.go
```diff
@@ -119,19 +121,120 @@ func (tm *Time) Excess() gas.Gas {
...
+// always returns minPrice regardless of excess. This prevents the fake
+// exponential approximation from producing e^(x/K) ≈ e^1 when excess is also
+// very large.
 func (tm *Time) Price() gas.Price {
-	return gas.CalculatePrice(1 /* M */, tm.excess, tm.excessScalingFactor())
+	if tm.config.staticPricing {
+		return tm.config.minPrice
+	}
+	return gas.CalculatePrice(tm.config.minPrice, tm.excess, tm.excessScalingFactor())
 }
```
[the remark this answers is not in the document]
ceyonur: I guess the problem here we're talking about if that happens the Price will be fixed at `MaxUint64` while excess increases so the recovery will be much harder.

If that's the case A) I don't think it's a very practical issue B) I think it will be fixed if we clamp excess to `MaxUint64`.

hook/hook.go
ceyonur: We absolutely can, but should we? That's more like a overall design/security question that lets them "bazooka their foot".

ceyonur: but I can't deny some people just want to watch the world burn...

Joke aside this can be a slightly viable option for TxAllowList enabled situation for L1s.

gastime/gastime_test.go
```diff
@@ -15,10 +15,18 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
+	"github.com/ava-labs/strevm/hook"
 	"github.com/ava-labs/strevm/intmath"
 	"github.com/ava-labs/strevm/proxytime"
 )
 
+func mustNew(tb testing.TB, at time.Time, target, startingExcess gas.Gas, gasConfig hook.GasConfig) *Time {
```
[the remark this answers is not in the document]
ceyonur: I think that's fine, otherwise we need to export it directly from `gastime` or change this file to be in `gastime_test` package (to prevent circular dependency)

hook/hook.go
```diff
@@ -41,6 +41,9 @@ type Points interface {
 	// GasTargetAfter returns the gas target that should go into effect
 	// immediately after the provided block.
 	GasTargetAfter(*types.Header) gas.Gas
+	// GasPriceConfigAfter returns the gas price configuration that should go into effect
+	// immediately after the provided block.
+	GasPriceConfigAfter(*types.Header) GasPriceConfig
```
ARR4N: Why is this separate from `GasTargetAfter()`? They're used sequentially in `gastime.Time.AfterBlock()` and take the identical argument. Why not something like the following?

```go
GasConfigAfter(*types.Header) GasConfig
```

```go
type GasConfig struct {
  // Target is ...
  Target gas.Gas
  // TargetToExcessScaling is ...
  TargetToExcessScaling gas.Gas
  // everything else currently in GasPriceConfig
}
```

intmath/intmath_test.go
```diff
@@ -32,6 +32,29 @@ func TestBoundedSubtract(t *testing.T) {
 	}
 }
 
+func TestBoundedMultiply(t *testing.T) {
+	tests := []struct {
+		a, b, max, want uint64
```
ARR4N:
```suggestion
		a, b, ceil, want uint64
```

intmath/intmath_test.go
```diff
@@ -32,6 +32,29 @@ func TestBoundedSubtract(t *testing.T) {
...
+		{a: 1, b: 1, max: 0, want: 0},               // max == 0 bounds everything
+		{a: 2, b: max, max: max, want: max},         // a*b would overflow uint64
+		{a: max, b: max, max: max, want: max},       // both at max, would overflow
+		{a: max, b: 2, max: max - 1, want: max - 1}, // a*b overflows, bounded to max-1
+	}
+
+	for _, tt := range tests {
+		if got := BoundedMultiply(tt.a, tt.b, tt.max); got != tt.want {
+			t.Errorf("BoundedMultiply[%T](%[1]d, %d, %d) got %d; want %d", tt.a, tt.b, tt.max, got, tt.want)
+		}
```
ARR4N:
```suggestion
		l, r := tt.a, tt.b
		for range 2 {
			if got := BoundedMultiply(l, r, tt.max); got != tt.want {
				t.Errorf("BoundedMultiply[%T](%[1]d, %d, %d) got %d; want %d", l, r, tt.max, got, tt.want)
			}
			l, r = r, l
		}
```

The function is commutative in the first arguments so we should test that property.

hook/hookstest/stub.go
```diff
@@ -21,23 +22,52 @@ import (
...
-	Ops    []hook.Op
+	Now            func() time.Time
+	Target         gas.Gas
+	GasPriceConfig hook.GasPriceConfig
+	Ops            []hook.Op
 }
 
 var _ hook.Points = (*Stub)(nil)
 
+// HookOption applies a configuration to [Stub].
```
ARR4N:
```suggestion
// A HookOption modifies the value returned by [NewStub].
```

hook/hookstest/stub.go
```diff
@@ -21,23 +22,52 @@ import (
...
+
+// WithGasPriceConfig overrides the default gas config.
+func WithGasPriceConfig(cfg hook.GasPriceConfig) HookOption {
+	return options.Func[Stub](func(s *Stub) {
+		s.GasPriceConfig = cfg
+	})
+}
+
+// WithNow overrides the default time source.
+func WithNow(now func() time.Time) HookOption {
```
ARR4N:
```suggestion
// WithNowFunc overrides the default time source.
func WithNowFunc(now func() time.Time) HookOption {
```

gastime/gastime.go
```diff
@@ -119,19 +121,105 @@ func (tm *Time) Excess() gas.Gas {
...
+	if err := newCfg.validate(); err != nil {
+		return err
+	}
+
+	currentPrice := tm.Price()
+	tm.config = newCfg
+	// Always call findExcessForPrice after updating the config for tm to ensure price continuity.
+	newExcess := tm.findExcessForPrice(currentPrice)
+	tm.excess = newExcess
+	return nil
```
ARR4N:
```suggestion
	p := tm.Price()
	tm.config = newCfg
	tm.excess = tm.findExcessForPrice(p)
	return nil
```

Rationale:
1. The initial-price variable is used over 3 lines so doesn't need to be so verbose, especially since its meaning can be inferred from the RHS ([see also](https://google.github.io/styleguide/go/decisions#variable-names)).
2. The `newExcess` variable was redundant.
3. You already named the method well so `tm.excess = findExcessForPrice(p)` is self-explanatory.

gastime/gastime.go
```diff
@@ -119,19 +121,105 @@ func (tm *Time) Excess() gas.Gas {
...
+// [config.targetToExcessScaling] * T, capped at [math.MaxUint64].
 func (tm *Time) excessScalingFactor() gas.Gas {
-	const overflowThreshold = math.MaxUint64 / TargetToExcessScaling
-	if tm.target > overflowThreshold {
-		return math.MaxUint64
+	return excessScalingFactorOf(tm.config.targetToExcessScaling, tm.target)
+}
+
+// excessScalingFactorOf returns scaling * target, capped at [math.MaxUint64].
+func excessScalingFactorOf(scaling, target gas.Gas) gas.Gas {
```
ARR4N: (subjective) Other than `Time.excessScalingFactor()` this is only used in a test where the context is sufficient to just use `intmath.BoundedMultiply()` directly. Usually I'm strongly in favour of coupling like that, but this also adds some unnecessary indirection (in `excessScalingFactor()`) and confusion (_What if the arguments are inverted? Ahh it doesn't matter because it's commutative._).

Overall I think the balance is in favour of removing it, but I appreciate that the argument is tenuous.

gastime/gastime.go
```diff
@@ -119,19 +121,105 @@ func (tm *Time) Excess() gas.Gas {
...
+	// We return 0 in case targetPrice < minPrice because we should at least maintain the minimum price
+	// by setting the excess to 0. ( P = M * e^(0 / K) = M )
+	// Note: Even though we return 0 for excess it won't avoid accumulating excess in the long run.
+	if targetPrice <= tm.config.minPrice || tm.config.staticPricing {
+		return 0
+	}
+
+	k := tm.excessScalingFactor()
+
+	// Binary search over [0, maxExcessSearchCap(k)] for smallest excess where price >= targetPrice.
```
ARR4N:
```suggestion
	// The price function is monotonic non-decreasing so binary search is appropriate.
```

The comment is outdated (`maxExcessSearchCap()` no longer exists) but it does warrant describing why binary search is safe.

gastime/acp176.go
```diff
@@ -21,13 +21,17 @@ func (tm *Time) BeforeBlock(hooks hook.Points, h *types.Header) {
...
 func (tm *Time) AfterBlock(used gas.Gas, hooks hook.Points, h *types.Header) error {
 	tm.Tick(used)
 	target := hooks.GasTargetAfter(h)
 	if err := tm.SetTarget(target); err != nil {
 		return fmt.Errorf("%T.SetTarget() after block: %w", tm, err)
 	}
+	cfg := hooks.GasPriceConfigAfter(h)
+	if err := tm.SetConfig(cfg); err != nil {
+		return fmt.Errorf("%T.SetConfig() after block: %w", tm, err)
+	}
```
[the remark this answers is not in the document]
ceyonur: I was not %100 sure that the `SetConfig` will be consumed in `AfterBlock` or in `BeforeBlock`. 
See https://github.com/ava-labs/strevm/pull/80#discussion_r2821477954

hook/hookstest/stub.go
```diff
@@ -21,23 +22,52 @@ import (
...
 // `Extra` field SHOULD NOT be modified as it encodes sub-second block time.
 func (s *Stub) BuildHeader(parent *types.Header) *types.Header {
-	var now time.Time
-	if s.Now != nil {
-		now = s.Now()
-	} else {
-		now = time.Now()
-	}
-
+	now := s.Now()
```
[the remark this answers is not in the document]
ceyonur: reverted all of these changes in stub file

gastime/gastime.go
```diff
@@ -50,16 +52,36 @@ func (tm *Time) establishInvariants() {
...
+		StaticPricing:         false,
+	}
+}
+
 // New returns a new [Time], derived from a [time.Time]. The consumption of
 // `target` * [TargetToRate] units of [gas.Gas] is equivalent to a tick of 1
-// second. Targets are clamped to [MaxTarget].
-func New(at time.Time, target, startingExcess gas.Gas) *Time {
+// second. Targets are clamped to [MaxTarget]. The gasPriceConfig parameter
+// specifies the gas pricing configurations.
```
ARR4N: (nit) The change in the comment is redundant, saying exactly what the code says.

gastime/gastime_test.go
```diff
@@ -336,27 +344,129 @@ func TestExcess(t *testing.T) {
...
+	t.Run("custom scaling", func(t *testing.T) {
+		for _, tt := range customTests {
+			tm := mustNew(t, time.Unix(0, 0), tt.target, 0, hook.GasPriceConfig{TargetToExcessScaling: tt.scaling, MinPrice: DefaultMinPrice})
+			require.NoErrorf(t, tm.SetTarget(tt.target), "%T.SetTarget(%v)", tm, tt.target)
+			assert.Equalf(t, tt.want, tm.excessScalingFactor(), "scaling=%d, T=%d", tt.scaling, tt.target)
+		}
+	})
+}
+
+func TestMinPrice(t *testing.T) {
```
ARR4N: These tests can be simplified and placed into a single table. The only thing that a test of `MinPrice()` needs to exercise is that it's used with and without static pricing. There's no need, for example to test different excess values because that's just a test of the `gas.CalculatePrice()` function from `avalanchego`. `SetConfig()` should be tested separately.

Price calculation is agnostic to the concept of a "default" so there's also no point in splitting out default and custom min prices when exercising the effects of this config value.

```go
func TestMinPrice(t *testing.T) {
	const (
		target = 1e6
		excess = target * DefaultTargetToExcessScaling // i.e. Price == floor(e*MinPrice)
	)

	tests := []struct {
		name     string
		minPrice gas.Price
		static   bool
		want     gas.Price
		wantErr  testerr.Want
	}{
		{
			name:     "min=1",
			minPrice: 1,
			want:     2,
		},
		{
			name:     "min=100",
			minPrice: 100,
			want:     271,
		},
		{
			name:     "high_min_no_overflow",
			minPrice: math.MaxUint64 / 2,
			want:     math.MaxUint64,
		},
		{
			name:     "zero_min_errors",
			minPrice: 0,
			wantErr:  testerr.Is(errMinPriceZero),
		},
		{
			name:     "static_pricing_returns_min",
			minPrice: 123_456,
			static:   true,
			want:     123_456,
		},
	}

	for _, tt := range tests {
		t.Run(tt.name, func(t *testing.T) {
			cfg := DefaultGasPriceConfig()
			cfg.MinPrice = tt.minPrice
			cfg.StaticPricing = tt.static

			tm, err := New(time.Unix(0, 0), target, excess, cfg)
			if diff := testerr.Diff(err, tt.wantErr); diff != "" {
				t.Fatalf("New(..., %+v) %s", cfg, diff)
			}
			if tt.wantErr != nil {
				return
			}
			if got := tm.Price(); got != tt.want {
				t.Errorf("New(..., excess=%d, %+v).Price() got %d; want %d", gas.Gas(excess), cfg, got, tt.want)
			}
		})
	}
}
```

ARR4N: Included in #214

ARR4N:
Please see #214 for simplification of the implementation and tests, with the tightest possible bounds on fuzzing.

PR #113, opened 2026-01-28
test: Add more `eth_` tests

alarso16:
A list of methods were put in #83 as implemented, but they weren't all tested. This adds some more tests to ensure coverage for each of these methods as well.

### DIFF
--- a/sae/rpc_test.go
+++ b/sae/rpc_test.go
@@ -223,15 +223,10 @@ func TestEthGetters(t *testing.T) {
 
 func testGetByHash(ctx context.Context, t *testing.T, sut *SUT, want *types.Block) {
 	t.Helper()
-	cmpOpts := []cmp.Option{
-		cmputils.Blocks(),
-		cmputils.Headers(),
-		cmpopts.EquateEmpty(),
-	}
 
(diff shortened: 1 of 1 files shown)


sae/rpc_test.go
```diff
@@ -217,6 +219,15 @@ func TestBlockGetters(t *testing.T) {
 							t.Errorf("Diff (-want +got):\n%s", diff)
 						}
 					})
+
+					txIndex := uint(i) //nolint:gosec // definitely won't overflow
+					hexInd := hexutil.Uint(txIndex)
```
ARR4N:
```suggestion
					hexInd := hexutil.Uint(i) //nolint:gosec // definitely won't overflow
```

No point going via `uint`. I think you can now call it `txIndex`, but adding that to the suggestion would break the rest of your code should you accept it.

ARR4N (dismissed):

alarso16:
> One additional comment and agree with Arran's comments, although have a preference for test tables.

The confusing part of the test is all the different calls, but since they all need to be different, I don't think it would be simpler to put in a table.

I think the invalid inputs need to be tested separately (at least a different PR), since we would frequently expect an error, and all the test helpers do not expect any error

PR #116, opened 2026-01-28
test: RPC resolution of pending block

ARR4N:
Extend tests of ava-labs/strevm#83 to include pending block.

### DIFF
--- a/sae/rpc_test.go
+++ b/sae/rpc_test.go
@@ -260,10 +260,17 @@ func TestTxPoolNamespace(t *testing.T) {
 	}...)
 }
 
-func TestBlockGetters(t *testing.T) {
+func TestEthGetters(t *testing.T) {
 	opt, vmTime := withVMTime(t, time.Unix(saeparams.TauSeconds, 0))
 
 	ctx, sut := newSUT(t, 1, opt)
+
+	t.Run("unknown_hashes", func(t *testing.T) {
+		sut.testGetByUnknownHash(ctx, t)
+	})
+	t.Run("unknown_numbers", func(t *testing.T) {
+		sut.testGetByUnknownNumber(ctx, t)
+	})
 
 	// The named block "pending" is the last to be enqueued but yet to be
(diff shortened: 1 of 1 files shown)


ARR4N:
Converting to draft, pending #119

sae/rpc_test.go
```diff
@@ -170,29 +173,49 @@ func TestEthGetters(t *testing.T) {
 	opt, vmTime := withVMTime(time.Unix(saeparams.TauSeconds, 0))
 
 	ctx, sut := newSUT(t, 1, opt)
-	genesis := sut.lastAcceptedBlock(t)
 
-	createTx := func(t *testing.T) *types.Transaction {
+	// The named block "pending" is the last to be enqueued but yet to be
+	// executed. Although unlikely to be useful in practice, it still needs to
+	// be tested.
+	blockingPrecompile := common.Address{'b', 'l', 'o', 'c', 'k'}
```
alarso16: This functionality works, but it does seem like we would want to generalize this to be able to easily add in multiple tests. Is that even possible? Can you temp register for the test and `t.Cleanup` the channel?

StephenButtolph (commented):
Was going to comment that we should probably add tests for:
1. Looking up pending blocks by hash
2. Looking up verified but not accepted blocks by hash (assuming we do want to support this, otherwise we should disallow it)

But this PR currently seems scoped to the `pending` block number... So I think it's fine

ARR4N:
> Was going to comment that we should probably add tests for:
> 
> 1. Looking up pending blocks by hash

We do, and by number. The `pending` variable is added to the loop immediately after its declaration, which means it's tested by both hash and explicit (not just named) number.

> 2. Looking up verified but not accepted blocks by hash (assuming we do want to support this, otherwise we should disallow it)
> 
> But this PR currently seems scoped to the `pending` block number... So I think it's fine

Tracking in https://github.com/ava-labs/strevm/issues/123

PR #117, opened 2026-01-28
test(rpc): add edge case tests and fix `GetPoolTransaction` panic

JonathanOppenheimer:
This PR adds edge case testing for the "`EthGetters`" Ethereum RPC methods and adds `GetPoolTransaction` to fix a panic when querying non-existent transactions (follow libevm flow to see why this is -- libevm ends up checking the mempool and oops we didn't look there kaboom). 

For edge case testing, I added `TestEthGettersEdgeCases` covering all the scenarios where Ethereum RPC methods should return null instead of errors:
                                                                                                                                                    
- Non-existent blocks/transactions - querying data that doesn't exist
- Future block numbers - blocks beyond the current chain height
- Out of bounds indices - transaction indices higher than the block's tx count
- Empty blocks - querying transactions from blocks with zero transactions

I looked up the Ethereum JSON-RPC specification (EIP-1474) to validate what the expected behavior was where missing data returns null, not errors. The functions comments may be a little excessive in this case, but I wanted to cite my sources. 

Lastly, I added `eth_getRawTransaction*` to the explicitly supported category. Should more testing of these in the happy path be done for these?

### DIFF
--- a/sae/rpc_test.go
+++ b/sae/rpc_test.go
@@ -230,6 +230,7 @@ func testGetByHash(ctx context.Context, t *testing.T, sut *SUT, want *types.Bloc
 	testRPCGetter(ctx, t, "eth_getBlockByHash", sut.BlockByHash, want.Hash(), want)
 	testRPCGetter(ctx, t, "eth_getBlockByHash", sut.HeaderByHash, want.Hash(), want.Header())
 	testRPCGetter(ctx, t, "eth_getBlockTransactionCountByHash", sut.TransactionCount, want.Hash(), uint(len(want.Transactions())))
+	testRPCMethod(ctx, t, sut, "eth_getUncleCountByBlockHash", hexutil.Uint(0), want.Hash())
 
 	for i, wantTx := range want.Transactions() {
 		hexInd := hexutil.Uint(i) //nolint:gosec // definitely won't overflow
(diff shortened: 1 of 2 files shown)


sae/rpc_test.go
```diff
@@ -219,49 +219,85 @@ func TestEthGetters(t *testing.T) {
...
 }
 
 func testGetByHash(ctx context.Context, t *testing.T, sut *SUT, want *types.Block) {
 	t.Helper()
 
-	testRPCGetter(ctx, t, "BlockByHash", sut.BlockByHash, want.Hash(), want)
-	testRPCGetter(ctx, t, "HeaderByHash", sut.HeaderByHash, want.Hash(), want.Header())
-	testRPCGetter(ctx, t, "TransactionCount", sut.TransactionCount, want.Hash(), uint(len(want.Transactions())))
+	testRPCGetter(ctx, t, "eth_getBlockByHash", sut.BlockByHash, want.Hash(), want)
+	testRPCGetter(ctx, t, "eth_getBlockByHash", sut.HeaderByHash, want.Hash(), want.Header())
```
ARR4N: These shouldn't have been changed. Although this is an indirect test of the underlying RPC methods, the error message uses them and will be misleading. I'm reverting them in #116.

PR #122, opened 2026-01-29
refactor: move invalid-block tests back to main testing function

StephenButtolph:
`TestBlockGetters` is confusing, as it tests _almost_ all of the cases for the block getters. It makes more sense to me to have that test all of the cases of the function, rather than splitting out some (not so special) edge cases to a different tester

### DIFF
--- a/sae/rpc_test.go
+++ b/sae/rpc_test.go
@@ -257,7 +257,7 @@ func TestTxPoolNamespace(t *testing.T) {
 	}...)
 }
 
-func TestBlockGetters(t *testing.T) {
+func TestEthGetters(t *testing.T) {
 	opt, vmTime := withVMTime(t, time.Unix(saeparams.TauSeconds, 0))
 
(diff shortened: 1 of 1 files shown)


sae/rpc_test.go
```diff
@@ -257,11 +257,17 @@ func TestTxPoolNamespace(t *testing.T) {
...
 
 	ctx, sut := newSUT(t, 1, opt)
-	genesis := sut.lastAcceptedBlock(t)
+
+	t.Run("unknown_hashes", func(t *testing.T) {
+		sut.testGetByUnknownHash(ctx, t)
+	})
+	t.Run("unknown_numbers", func(t *testing.T) {
+		sut.testGetByUnknownNumber(ctx, t)
+	})
```
ARR4N: I'm much happier with these being up here and immediately after `newSUT()` so they clearly signal "you don't need to remember anything".

PR #125, opened 2026-01-30
feat: Minimal support of `eth_` RPCs depending on the keystore

JonathanOppenheimer:
This PR adds support for `eth_sendTransaction`,`eth_sign`, `eth_signTransaction`, and `eth_pendingTransactions` RPC methods for Ethereum API compatibility. I added explicit tests for all, and added them to the list.

Note that the expected behavior for some of these tests is an unknown account error, as strevm will not configure keystore backends by design since we don't store private keys on nodes.

### DIFF
--- a/saexec/execution.go
+++ b/saexec/execution.go
@@ -27,32 +27,30 @@ var errExecutorClosed = errors.New("saexec.Executor closed")
 // before [blocks.Block.Executed] returns true then there is no guarantee that
 // the block will be executed.
 func (e *Executor) Enqueue(ctx context.Context, block *blocks.Block) error {
-	warnAfter := time.Millisecond
-	for {
-		select {
-		case e.queue <- block:
-			e.lastEnqueued.Store(block)
-			e.enqueueEvents.Send(block.EthBlock())
-			return nil
+	select {
+	case e.queue <- block:
+		e.lastEnqueued.Store(block)
+		e.enqueueEvents.Send(block.EthBlock())
 
-		case <-ctx.Done():
-			return ctx.Err()
-
-		case <-e.quit:
-			return errExecutorClosed
-		case <-e.done:
-			// `e.done` can also close due to [Executor.execute] errors.
-			return errExecutorClosed
-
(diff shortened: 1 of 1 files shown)


sae/rpc_test.go
```diff
@@ -353,6 +353,93 @@ func TestEthGetters(t *testing.T) {
...
+		require.NoError(t, err)
+		require.Empty(t, pendingTxs, "pendingTransactions filters to known accounts only")
+	})
+
+	t.Run("eth_sendTransaction", func(t *testing.T) {
+		ctx, sut := newSUT(t, 1)
+
+		// eth_sendTransaction should fail with "unknown account" since no keystore is configured.
+		// This is intended behavior and the extent of our 'support' of this method - we do not
+		// want to store private keys on the node.
```
JonathanOppenheimer: Confirmed to me by @StephenButtolph

sae/rpc.go
```diff
@@ -84,11 +90,16 @@ func (vm *VM) ethRPCServer() (*rpc.Server, error) {
 		// - eth_getTransactionByBlockHashAndIndex
 		// - eth_getTransactionByBlockNumberAndIndex
 		// - eth_getTransactionByHash
+		// - eth_sendRawTransaction
```
ARR4N: This is already supported; it's how `SUT.mustSendTx()` works.

sae/rpc.go
```diff
@@ -32,9 +34,13 @@ import (
 
 // APIBackend returns an API backend backed by the VM.
 func (vm *VM) APIBackend() ethapi.Backend {
+	accountManager := accounts.NewManager(&accounts.Config{})
+	vm.toClose = append(vm.toClose, accountManager.Close)
```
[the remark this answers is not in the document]
JonathanOppenheimer: This feels slightly out of scope for this PR, but it is new behavior introduced here! 

https://github.com/ava-labs/strevm/pull/125/changes/8f03c5cfee2adddfe56f4b513b5a79c6eb72624c

PR #130, opened 2026-02-02
chore: attempt to disable Copilot

ARR4N:
In the absence of an explicit setting to just disable it (because in the age of AI, everything must be fuzzy logic). Motivated by #129.


JonathanOppenheimer:
Does this series of instructions not work https://github.com/orgs/community/discussions/180848#discussioncomment-15127348? I can see the option in one of the repositories I own (outside of Ava Labs)

PR #132, opened 2026-02-02
test: support `eth_chainId`

JonathanOppenheimer:
Another free win, just a bit of test coverage and it's in the list!

### DIFF
--- a/sae/rpc_test.go
+++ b/sae/rpc_test.go
@@ -21,6 +21,7 @@ import (
 	"github.com/ava-labs/libevm/libevm"
 	"github.com/ava-labs/libevm/libevm/ethapi"
 	libevmhookstest "github.com/ava-labs/libevm/libevm/hookstest"
+	"github.com/ava-labs/libevm/libevm/options"
 	"github.com/ava-labs/libevm/params"
 	"github.com/ava-labs/libevm/rpc"
 	"github.com/google/go-cmp/cmp"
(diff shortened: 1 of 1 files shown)


ARR4N:
I don't know why CI wasn't running, so I clicked the `Update Branch` button to get the merge to trigger it.

sae/rpc_test.go
```diff
@@ -260,6 +260,14 @@ func TestTxPoolNamespace(t *testing.T) {
 	}...)
 }
 
+func TestChainID(t *testing.T) {
+	ctx, sut := newSUT(t, 1)
+	sut.testRPC(ctx, t, rpcTest{
+		method: "eth_chainId",
+		want:   hexutil.Uint64(saetest.ChainConfig().ChainID.Uint64()),
+	})
+}
```
ARR4N:
```suggestion
for id := range uint64(2) {
		ctx, sut := newSUT(t, 0, options.Func[sutConfig](func(c *sutConfig) {
			c.genesis.Config = &params.ChainConfig{
				ChainID: new(big.Int).SetUint64(id),
			}
		}))
		sut.testRPC(ctx, t, rpcTest{
			method: "eth_chainId",
			want:   hexutil.Uint64(id),
		})
	}
}
```

Rationale:

1. The original approach relies on internal knowledge of `newSUT()` (i.e. that it uses `saetest.ChainConfig()`. This might [change](https://testing.googleblog.com/2015/01/testing-on-toilet-change-detector-tests.html), and also stops the test code from being self-evidently correct.
2. One property of a good test is that it exercises/demonstrates [behaviour](https://testing.googleblog.com/2013/08/testing-on-toilet-test-behavior-not.html) of the code, which is almost always "if I put `x` in, I get `y` out" (in this case `x == y`) or some more convoluted version of the same.

Issue #127, opened 2026-01-30
Fix `TestWorstCase` flake


JonathanOppenheimer:
See https://github.com/ava-labs/strevm/actions/runs/21533722194/job/62055016856?pr=126

Issue #137, opened 2026-02-04
Fix `TestContextualOpCodes` flake


JonathanOppenheimer:
See https://github.com/ava-labs/strevm/actions/runs/21682846439/job/62521812244?pr=124

PR #138, opened 2026-02-04
fix: all `execution queue buffer too small` flakes

JonathanOppenheimer:
Closes https://github.com/ava-labs/strevm/issues/127, #137, #145. Fixes test flakes caused by misleading "Execution queue buffer too small" warnings on CPU-constrained CI runners.

For context, fuzz tests occasionally fails in CI when the execution queue takes >1ms to accept a block:, 

 ```go
  case <-time.After(warnAfter):
      e.log.Warn("Execution queue buffer too small", ...)
````

triggering a "Execution queue buffer too small" warning. (I've seen this a few times -- didn't save all the CI runs, but you can see it happens on at least 3 different test cases). Since the test logger treats WARN as test failure, this causes flakes. 

This warning doesn't actually check if the buffer is full - it only measures latency. The warning fires in CPU constrained environments, when it shouldn't. By actually checking whether the buffer is close to full, we can avoid this. 80% seems reasonable but I can change it.

### DIFF
--- a/saexec/execution.go
+++ b/saexec/execution.go
@@ -34,7 +34,7 @@ func (e *Executor) Enqueue(ctx context.Context, block *blocks.Block) error {
 
 		size := cap(e.queue)
 		warningThreshold := size - size/16
-		if n := len(e.queue); n > warningThreshold {
+		if n := len(e.queue); n >= warningThreshold {
 			// If this happens then increase the channel's buffer size.
 			e.log.Warn(
 				"Execution queue buffer too small",

### CODE REVIEW

**Comment by JonathanOppenheimer** [comment]
> A few comments:
> 
> 1. I don't think this PR does anything.
> 
> ```go
> // NewTBLogger constructs a logger that propagates logs to [testing.TB]. WARNING
> // and ERROR logs are sent to [testing.TB.Errorf] while FATAL is sent to
> // [testing.TB.Fatalf]. All other logs are sent to [testing.TB.Logf]. Although
> // the level can be configured, it is silently capped at [logging.Warn].
> ```
> 
> 2. While we may need to make exceptions around the `warnings == errors` at some point, we should probably be making that argument on _new_ code rather than code written to expect that `warnings == errors`.
> 
> 3. I don't think these are fair to categorize as "just a testing flake". The executor's buffer is initialized to 4k blocks, which is more than we put in the queue during these tests. These flakes are showing that `Execution queue buffer too small` is an invalid log (in CPU constrained environments).

Ahhh.... for `Enqueue` 

```go 
case <-time.After(warnAfter):
              e.log.Warn("Execution queue buffer too small", ...) 
              warnAfter *= 2
          }
``` 

triggers because of the very resource constrained environment in CI (I assume the time just elapses bc the go routine is starving). The means the log message doesn't really make sense. I should have been more critical about thinking. We can modify this to check if the buffer actually is full? That makes the most sense to me.

ARR4N:
I think 78ba34bdaeac0ca4dbbab44a0ffbb3c5835c7629 addresses the root of the problem.

saexec/execution.go
```diff
@@ -27,12 +27,24 @@ var errExecutorClosed = errors.New("saexec.Executor closed")
...
-	warnAfter := time.Millisecond
 	for {
+		size := cap(e.queue)
+
 		select {
 		case e.queue <- block:
 			e.lastEnqueued.Store(block)
 			e.enqueueEvents.Send(block.EthBlock())
+
+			if n := len(e.queue); n > size-size/16 {
```
StephenButtolph: Can we make `size - size/16` a variable called `warningThreshold` or something?

PR #139, opened 2026-02-04
test: fix `TestAcceptBlock` flake

### LINKED ISSUE #133: Fix `TestAcceptBlock` flake
See https://github.com/ava-labs/strevm/actions/runs/21603826370/job/62256297607

JonathanOppenheimer:
Closes #133. 

Occasionally this test will fail with 

```
Error: Should be zero, but was 2
Messages: initial in-memory block count
```

Under CI,  previous tests' finalizers can take >100ms to run, which does not give the GC adequate time to run -- this gives the GC ample time.

### DIFF
--- a/sae/rpc.go
+++ b/sae/rpc.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/ava-labs/avalanchego/network/p2p"
 	"github.com/ava-labs/avalanchego/version"
+	ethereum "github.com/ava-labs/libevm"
 	"github.com/ava-labs/libevm/common"
 	"github.com/ava-labs/libevm/common/hexutil"
 	"github.com/ava-labs/libevm/core"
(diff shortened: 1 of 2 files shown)


sae/vm_test.go
```diff
@@ -547,10 +547,14 @@ func TestSyntacticBlockChecks(t *testing.T) {
 }
 
 func TestAcceptBlock(t *testing.T) {
+	// Wait for GC finalizers from previous tests to run. Blocks are tracked via
+	// runtime.AddCleanup, which runs asynchronously during GC. In CI environments,
+	// finalizers can take longer to run, so we use a generous timeout to ensure
+	// we start with a clean block count. Zero!
```
StephenButtolph:
```suggestion
	// In resource constrained environments, finalizers can take a long time to
	// run, so we need to use a generous timeout.
```

PR #141, opened 2026-02-04
test: remove exact string match from `txpool_*` tests

JonathanOppenheimer:
Closes ava-labs/avalanchego#5270 

This test was copy-pasting libevm's transaction formatting logic to check `txpool_*` output, which is in fact pretty gross. Part of it also wasn't even correct as we used `GasFeeCap()` instead of `GasPrice()`. It also tested libevm's formatting instead of our actual backend (change detector tests bad! -arran). 

Instead of copying the formatter, we can call libevm's functions directly and use that as the expected result.

### DIFF
--- a/sae/rpc_test.go
+++ b/sae/rpc_test.go
@@ -179,59 +179,36 @@ func TestTxPoolNamespace(t *testing.T) {
 		queuedAccount  = 1
 	)
 	pendingTx := makeTx(pendingAccount)
-	pendingRPCTx := ethapi.NewRPCPendingTransaction(pendingTx, nil, saetest.ChainConfig())
-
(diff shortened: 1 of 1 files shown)


StephenButtolph:
This PR would result in testing the code against itself. Ensuring that the API provides consistent results... Not necessarily correct results.

If we could somehow use the standard ethereum API backend to perform this testing, then it could maybe make sense... But that's not what we're doing here (and I don't think it is easy to do that here).

I'm going to close this PR... Feel free to make the more specific change of `tx.GasFeeCap().Uint64()` into `tx.GasPrice()`. Technically `tx.GasPrice() == tx.GasFeeCap()` for all the existing transactions... But that isn't necessarily the case going forward.

PR #142, opened 2026-02-05
Simplify keystore tests

StephenButtolph:


sae/rpc.go
```diff
@@ -252,10 +252,6 @@ func (b *ethAPIBackend) GetPoolTransactions() (types.Transactions, error) {
 	return txs, nil
 }
 
-func (b *ethAPIBackend) GetPoolNonce(ctx context.Context, addr common.Address) (uint64, error) {
-	return b.Set.Pool.Nonce(addr), nil
-}
-
```
StephenButtolph: None of the added RPCs can hit this function - so adding it here would leave it untested.

PR #149, opened 2026-02-05
feat: Add `ParseGenesis` as initialization hook

michaelkaplan13:
Adds a `ParseGenesis` hook to allow for custom genesis initialization defined by the consumer.

This allows the consumer to initialize EVM upgrade activation times, as discussed in this [comment](https://github.com/ava-labs/strevm/pull/81#issuecomment-3797074015). I was able to confirm on the AvalancheGo side that providing a `coreth` compatible `ParseGenesis` [implementation](https://github.com/ava-labs/avalanchego/blob/50d638bb02d18ac64bcec0565b86b2a9f756ed1d/vms/saevm/hooks.go#L81) results in newer transaction types being support upon initialization of the SAE VM.

### DIFF
--- a/hook/hook.go
+++ b/hook/hook.go
@@ -12,7 +12,7 @@ import (
 	"time"
 
 	"github.com/ava-labs/avalanchego/ids"
-	"github.com/ava-labs/avalanchego/snow"
+	"github.com/ava-labs/avalanchego/upgrade"
 	"github.com/ava-labs/avalanchego/vms/components/gas"
 	"github.com/ava-labs/libevm/common"
 	"github.com/ava-labs/libevm/core"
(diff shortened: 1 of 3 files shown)


ARR4N:
What kind of changes does the hook make after JSON decoding? It feels a bit weird that the genesis bytes don't include sufficient information to be unmarshalled via the standard means.

EDIT: sorry, I missed that you'd linked to the implementation. libevm extras support JSON round-tripping so why aren't they just encoded into the genesis bytes?

michaelkaplan13:
> What kind of changes does the hook make after JSON decoding? It feels a bit weird that the genesis bytes don't include sufficient information to be unmarshalled via the standard means.
> 
> EDIT: sorry, I missed that you'd linked to the implementation. libevm extras support JSON round-tripping so why aren't they just encoded into the genesis bytes?

The honest answer is because I wasn't aware/didn't think of that. 🙂 

Would this work for supporting chains though? As far as I understand, the Avalanche upgrade activation times are not encoded into the C-Chain or EVM L1 genesis blocks, and instead provided from the AvalancheGo node level configuration through the `snowCtx` to individual chains. We could also encode the Avalanche upgrade activations in the genesis bytes of new chains, but I don't think we can add it to existing chains?

ARR4N:
> As far as I understand, the Avalanche upgrade activation times are not encoded into the C-Chain or EVM L1 genesis blocks, and instead provided from the AvalancheGo node level configuration through the `snowCtx` to individual chains.

That was my understanding too. So `SinceGenesis.Initialize()` needs to handle `snowCtx.NetworkUpgrades`. It might still require a hook like `ApplyNetworkUpgrades(*params.ChainConfig, []foo.NetworkUpgrade) error` though.

PR #150, opened 2026-02-05
feat: `eth_getTransactionReceipt` and `eth_getBlockReceipts` RPCs

### LINKED ISSUE #91: API Support - Receipt retrieval
Get transaction receipts for a block.

**Methods**:
- `GetReceipts()`

JonathanOppenheimer:
Closes #91

  This PR adds RPC support for:
  - `eth_getTransactionReceipt`
  - `eth_getBlockReceipts`

## Main Changes

- Overrode `BlockChainAPI`'s `GetBlockReceipts` to correctly handle queries for unexecuted blocks in `eth_getBlockReceipts`.
- Implemented `GetReceipts(ctx, hash)`
- Implemented `BlockByNumberOrHash(ctx, blockNrOrHash)`
- Implemented `HeaderByNumberOrHash(ctx, blockNrOrHash)`

To ensure RPC receipts are consistent across in-memory and restored paths:

- `saexec/execution.go`
  - Populates `receipt.EffectiveGasPrice` during execution
- `blocks/execution.go`
  - `RestoreExecutionArtefacts(...)` now derives receipt fields from raw receipts

### DIFF
--- a/sae/rpc_test.go
+++ b/sae/rpc_test.go
@@ -720,7 +720,6 @@ func TestGetReceipts(t *testing.T) {
 	unsettled, wantUnsettled := slice(t, 4, 6)
 	sut.waitUntilExecuted(t, unsettled)
 
-	// Pending block: accepted but not yet executed (blocking precompile).
 	pendingTx := sut.wallet.SetNonceAndSign(t, 0, &types.LegacyTx{
 		To:       &blockingPrecompile,
 		Gas:      params.TxGas,


sae/rpc_test.go
```diff
@@ -378,6 +379,189 @@ func TestEthGetters(t *testing.T) {
 	})
 }
 
+func TestReceiptAPIs(t *testing.T) {
+	opt, vmTime := withVMTime(t, time.Unix(saeparams.TauSeconds, 0))
+	ctx, sut := newSUT(t, 5, opt)
```
JonathanOppenheimer: Creating the `sut` here is acceptable from my understanding. We create and use 5 accounts to avoid logging logging nonce stuff.

sae/rpc.go
```diff
@@ -318,6 +348,67 @@ func (b *ethAPIBackend) SubscribePendingLogsEvent(chan<- []*types.Log) event.Sub
...
+		return nil, nil
+	}
+
+	header := rawdb.ReadHeader(b.vm.db, hash, number)
+	if header == nil {
+		return nil, nil
+	}
+
+	// Compute blobGasPrice the same way the executor does (saexec/execution.go)
+	// to keep the DB path consistent with the cache path for blob transactions.
```
JonathanOppenheimer: I think this is a reasonable "why" comment

sae/rpc_test.go
```diff
@@ -378,6 +380,198 @@ func TestEthGetters(t *testing.T) {
...
+		return sut.wallet.SetNonceAndSign(t, account, &types.DynamicFeeTx{
+			To:        &to,
+			Gas:       params.TxGas,
+			GasFeeCap: big.NewInt(1),
+		})
+	}
+	getReceipt := func(t *testing.T, tx *types.Transaction) *types.Receipt {
+		t.Helper()
+		var receipt *types.Receipt
+		require.NoError(t, sut.CallContext(ctx, &receipt, "eth_getTransactionReceipt", tx.Hash()))
```
JonathanOppenheimer: We use call context because we need the receipt.

JonathanOppenheimer:
Question for @StephenButtolph -- I know you want to avoid the use of `CallContext` and use the RPC helper methods instead -- here I found it unavoidable. Would you like another RPC helper method which will do the call, and return something?

sae/rpc_test.go
```diff
@@ -379,6 +381,197 @@ func TestEthGetters(t *testing.T) {
...
+		return receipt
+	}
+
+	var zeroAddr common.Address
+
+	// All blocks created upfront to avoid executor blocking during subtests.
+
+	genesis := sut.lastAcceptedBlock(t)
+
+	// Block 1: Executed, still in cache (not yet settled)
```
alarso16: These comments don't make sense. Since block 2 is settled, this one will also be settled (if you don't believe me, you should check)

sae/rpc_test.go
```diff
@@ -379,6 +381,197 @@ func TestEthGetters(t *testing.T) {
...
+	txExecutedInCache := createTx(t, 0, zeroAddr)
+	blockExecutedInCache := sut.createAndAcceptBlock(t, txExecutedInCache)
+	require.NoErrorf(t, blockExecutedInCache.WaitUntilExecuted(ctx), "%T.WaitUntilExecuted()", blockExecutedInCache)
+
+	// Block 2: Will be settled to DB
```
alarso16: This one won't be in the DB yet. All parents of the this block will be though

sae/rpc_test.go
```diff
@@ -378,6 +380,198 @@ func TestEthGetters(t *testing.T) {
...
+		} {
+			t.Run(tc.name, func(t *testing.T) {
+				requireBlockReceipts(t, tc.id, tc.txs)
+			})
+		}
+
+		t.Run("hash_equals_number", func(t *testing.T) {
+			var byHash, byNum []*types.Receipt
+			require.NoError(t, sut.CallContext(ctx, &byHash, "eth_getBlockReceipts", blockMultiTxs.Hash()))
+			require.NoError(t, sut.CallContext(ctx, &byNum, "eth_getBlockReceipts", hexutil.Uint64(blockMultiTxs.Height())))
```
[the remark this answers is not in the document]
alarso16: For what?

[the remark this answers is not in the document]
JonathanOppenheimer: This test is now redundant and can be deleted entirely.

saexec/saexec_test.go
```diff
@@ -313,7 +313,7 @@ func TestExecution(t *testing.T) {
 	opts := cmp.Options{
 		cmpopts.IgnoreFields(
 			types.Receipt{},
-			"GasUsed", "CumulativeGasUsed",
+			"GasUsed", "CumulativeGasUsed", "EffectiveGasPrice",
```
ARR4N: Why ignore this now? Proper gas accounting is an extremely important part of the `saexec.Executor`. If the original values were wrong, it's better to populate them correctly now than to simply ignore them.

(fyi) Having to change any existing test is usually smoke indicating a potential fire. Having to change one completely unrelated to the intent of the PR should be considered a massive red flag. It's not that it can't be done, but it needs to be explained.

saexec/execution.go
```diff
@@ -155,15 +157,6 @@ func (e *Executor) execute(b *blocks.Block, logger logging.Logger) error {
 		// the queue. It's only worth it if [blocks.LastToSettleAt] regularly
 		// returns false, meaning that execution is blocking consensus.
 
-		// The [types.Header] that we pass to [core.ApplyTransaction] is
-		// modified to reduce gas price from the worst-case value agreed by
-		// consensus. This changes the hash, which is what is copied to receipts
-		// and logs.
-		receipt.BlockHash = b.Hash()
```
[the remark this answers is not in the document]
StephenButtolph: fwiw, the `DeriveFields` call _does_ set this correctly: https://github.com/ava-labs/libevm/blob/6804c3c0f63b79080e18f4c7c42b70d62f250afe/core/types/receipt.go#L346

The bigger concern here is that the `receipt` is missing the `EffectiveGasPrice` value after being returned by `core.ApplyTransaction`... TBH this might be a geth bug...

If we wanted to correctly set the `receipt` here, and avoid calling `DeriveFields` below, we would only need to set the `EffectiveGasPrice` correctly.

sae/rpc.go
```diff
@@ -293,17 +295,49 @@ func (b *ethAPIBackend) BlockByNumber(ctx context.Context, n rpc.BlockNumber) (*
 }
 
 func (b *ethAPIBackend) HeaderByHash(ctx context.Context, hash common.Hash) (*types.Header, error) {
-	if b, ok := b.vm.blocks.Load(hash); ok {
-		return b.Header(), nil
+	return b.headerByHash(ctx, hash, false)
+}
+
+func (b *ethAPIBackend) headerByHash(_ context.Context, hash common.Hash, canonical bool) (*types.Header, error) {
+	// In-memory cache blocks are always canonical (SAE never reorgs).
```
JonathanOppenheimer: sanity check?

sae/rpc.go
```diff
@@ -293,17 +295,49 @@ func (b *ethAPIBackend) BlockByNumber(ctx context.Context, n rpc.BlockNumber) (*
...
+	num := rawdb.ReadHeaderNumber(b.vm.db, hash)
+	if num == nil {
+		return nil, nil
 	}
-	return readByHash(b, hash, rawdb.ReadBlock), nil
+	if canonical && rawdb.ReadCanonicalHash(b.vm.db, *num) != hash {
+		return nil, errHashNotCanonical
+	}
+	return rawdb.ReadBlock(b.vm.db, hash, *num), nil
+}
```
JonathanOppenheimer: These functions are nearly identical. Would the following be preferred? 

```go
// lookupByHash checks the in-memory cache first (all cached blocks are
// canonical because SAE never reorgs), then falls back to the database.
func lookupByHash[T any](
    b *ethAPIBackend,
    hash common.Hash,
    canonical bool,
    fromCache func(*blocks.Block) *T,
    fromDB canonicalReader[T],
) (*T, error) {
    if blk, ok := b.vm.blocks.Load(hash); ok {
        return fromCache(blk), nil
    }

    num := rawdb.ReadHeaderNumber(b.vm.db, hash)
    if num == nil {
        return nil, nil
    }

    if canonical && rawdb.ReadCanonicalHash(b.vm.db, *num) != hash {
        return nil, errHashNotCanonical
    }

    return fromDB(b.vm.db, hash, *num), nil
}
```

saexec/execution.go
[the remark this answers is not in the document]
JonathanOppenheimer: See current state now.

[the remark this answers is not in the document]
JonathanOppenheimer: It is _extremely weird_ that geth doesn't do this. I am inclined to agree it's a bug.

sae/vm.go
```diff
@@ -147,7 +150,14 @@ func NewVM(
 		}
 	}
 
-	rec := &recovery{db, xdb, snowCtx.Log, cfg, lastSync}
+	rec := &recovery{
```
ARR4N: Please revert it and just put the `xdb` in the right place. As an internal struct it's ok to not name the fields, and it has the added benefit of the compiler alerting you when one is missing. One caveat of this approach is that fields with the same type can be a gotcha (not the case here).

saexec/execution.go
```diff
@@ -154,10 +155,18 @@ func (e *Executor) execute(b *blocks.Block, logger logging.Logger) error {
 		// modified to reduce gas price from the worst-case value agreed by
 		// consensus. This changes the hash, which is what is copied to receipts
 		// and logs.
+		//
+		// [core.ApplyTransaction] also doesn't set EffectiveGasPrice for some
+		// reason. Fix both here so cached receipts are correct without a
+		// DeriveFields pass.
```
ARR4N:
```suggestion
		// [types.Receipt.DeriveFields] pass.
```

sae/rpc.go
```diff
@@ -293,17 +296,58 @@ func (b *ethAPIBackend) BlockByNumber(ctx context.Context, n rpc.BlockNumber) (*
...
+func (b *ethAPIBackend) BlockByNumberOrHash(ctx context.Context, blockNrOrHash rpc.BlockNumberOrHash) (*types.Block, error) {
+	block, err := readByNumberOrHash(ctx, blockNrOrHash, b.BlockByNumber, b.blockByHash)
+	if block == nil || err != nil {
+		return block, err
+	}
+	if inMemory, ok := b.vm.blocks.Load(block.Hash()); ok && !inMemory.Executed() {
+		// Block receipts/state are unavailable until execution completes.
+		return nil, nil
+	}
+	return block, nil
```
ARR4N:
```suggestion
	return readByNumberOrHash(ctx, blockNrOrHash, b.BlockByNumber, b.blockByHash)
```

This method has nothing to do with receipts and state so the check for execution status is misplaced here.

sae/rpc.go
```diff
@@ -405,6 +445,23 @@ func (b *ethAPIBackend) resolveBlockNumber(bn rpc.BlockNumber) (uint64, error) {
...
 
+func readByNumberOrHash[T any](
+	ctx context.Context,
+	blockNrOrHash rpc.BlockNumberOrHash,
+	byNum func(context.Context, rpc.BlockNumber) (*T, error),
+	byHash func(context.Context, common.Hash, bool) (*T, error),
+) (*T, error) {
+	if n, ok := blockNrOrHash.Number(); ok {
+		// Lookup by number is always canonical because readByNumber resolves
+		// the hash via rawdb.ReadCanonicalHash.
```
ARR4N: This is brittle. A future change to the implementation doesn't know anything about this assumption and could break it.

There is also no guarantee that `byNum` is actually carrying `readByNum`. That said, why are `byNum` and `byHash` received as arguments? This code is so general-purpose that it's very hard to reason about.

sae/rpc_test.go
```diff
@@ -654,6 +656,240 @@ func TestEthPendingTransactions(t *testing.T) {
 	})
 }
 
+func TestReceiptAPIs(t *testing.T) {
```
ARR4N: (note to self) Not reviewed yet.

sae/rpc_test.go
```diff
@@ -654,6 +656,240 @@ func TestEthPendingTransactions(t *testing.T) {
...
+	opt, vmTime := withVMTime(t, time.Unix(saeparams.TauSeconds, 0))
+	ctx, sut := newSUT(t, 5, opt)
+
+	// Blocking precompile creates accepted-but-not-executed blocks
```
ARR4N: Please don't copy-paste code that should be abstracted and reused. Once #179 is merged, please use it here.

sae/rpc.go
```diff
@@ -293,17 +295,53 @@ func (b *ethAPIBackend) BlockByNumber(ctx context.Context, n rpc.BlockNumber) (*
...
+	}
+	num, hash, err := b.resolveBlockNumberOrHash(blockNrOrHash)
+	if err != nil {
+		return nil, err
+	}
+	if blk, ok := b.vm.blocks.Load(hash); ok {
+		return blk.Header(), nil
 	}
-	return readByHash(b, hash, rawdb.ReadBlock), nil
+	return rawdb.ReadHeader(b.vm.db, hash, num), nil
```
JonathanOppenheimer: I got rid all of the indirection as you suggested. Easier to read, but more duplicated code, with just the type differing. Is this preferred?

JonathanOppenheimer: okay the helper is back lol

StephenButtolph (commented):
This PR LGTM... Since I've made some substantial modifications I'll defer to @ARR4N to actually approve it.

ARR4N:
I'm not gonna merge because I'm not sure if the PR description is up to date with the changes. If/when it's in line, please go ahead @JonathanOppenheimer.

StephenButtolph:
I updated the description

PR #151, opened 2026-02-07
feat: support millisecond block times

### LINKED ISSUE #49: Support millisecond block times in `blocks.LastToSettleAt()`.

JonathanOppenheimer:
Closes #49 -- adds `TimestampMilliseconds` to `hook.Points` (why is it called Points by the way?), giving callers direct access to the ACP-226 millisecond timestamp. Do note that any existing uses of the hook will need to be updated.  `PreciseTime` now uses `TimestampMilliseconds` instead of reconstructing the time from `seconds + SubSecondBlockTime`. 

I also added 2 new tests, one to test pre ACP-226 timing, and `TestLastToSettleAtMilliseconds` to test settlement of blocks within the same second (pretty cool functionality!).

### DIFF
--- a/hook/hookstest/stub.go
+++ b/hook/hookstest/stub.go
@@ -41,8 +41,8 @@ func (s *Stub) BuildHeader(parent *types.Header) *types.Header {
 	hdr := &types.Header{
 		ParentHash: parent.Hash(),
 		Number:     new(big.Int).Add(parent.Number, common.Big1),
-		Time:       uint64(now.Unix()),                                           //nolint:gosec // Known non-negative
-		Extra:      binary.BigEndian.AppendUint64(nil, uint64(now.UnixMilli())),  //nolint:gosec // Known non-negative
+		Time:       uint64(now.Unix()),                                          //nolint:gosec // Known non-negative
+		Extra:      binary.BigEndian.AppendUint64(nil, uint64(now.UnixMilli())), //nolint:gosec // Known non-negative
 	}
 	return hdr
 }


blocks/time_test.go
```diff
@@ -60,15 +73,22 @@ func FuzzTimeExtraction(f *testing.F) {
...
+	// ACP-226 provides millisecond precision, so the fuzz input is in
+	// milliseconds (0-999) rather than arbitrary nanoseconds.
 
-	f.Fuzz(func(t *testing.T, unix int64, target uint64, subSec int64) {
-		if subSec < 0 || time.Duration(subSec) >= time.Second {
-			t.Skip("Invalid sub-second value")
+	f.Fuzz(func(t *testing.T, unix int64, target uint64, subSecMs int64) {
+		if unix < 0 {
+			t.Skip("ACP-226 timestamps are non-negative")
+		}
```
JonathanOppenheimer: I am pretty confident we can make this assertion.

blocks/time_test.go
```diff
@@ -52,6 +54,17 @@ func TestGasTime(t *testing.T) {
...
+func TestPreciseTime_PreACP226Fallback(t *testing.T) {
+	hooks := &hookstest.Stub{}
+	hdr := &types.Header{
+		Time: 42,
+	}
+
+	got := PreciseTime(hooks, hdr)
+	want := time.UnixMilli(42_000)
+	require.Equal(t, want, got)
+}
```
JonathanOppenheimer: reasonable I think

sae/vm_test.go
```diff
@@ -207,15 +207,15 @@ func (t *vmTime) now() time.Time {
 }
 
 func (t *vmTime) set(n time.Time) {
-	t.Time = n
+	t.Time = n.Truncate(time.Millisecond)
```
ARR4N: The ability to achieve the same result by truncating an existing timestamp demonstrates that millisecond resolution was already possible.

PR #157, opened 2026-02-10
test: Allow error-checking

alarso16:
Credit to @powerslider for the implementation here, but thought we should pull it out from #154, since it can be used elsewhere, and can easily be isolated.

### DIFF
--- a/sae/rpc_test.go
+++ b/sae/rpc_test.go
@@ -30,6 +30,7 @@ import (
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/holiman/uint256"
+	"github.com/mrwormhole/errdiff"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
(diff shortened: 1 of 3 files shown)


powerslider (approved):

sae/rpc_test.go
```diff
@@ -486,6 +501,7 @@ func TestEthPendingTransactions(t *testing.T) {
 func TestEthSigningAPIs(t *testing.T) {
 	ctx, sut := newSUT(t, 1)
 
+	wantErr := "unknown account"
```
StephenButtolph: pedantic:
```suggestion
	const wantErr = "unknown account"
```

sae/rpc_test.go
```diff
@@ -55,11 +56,25 @@ func (s *SUT) testRPC(ctx context.Context, t *testing.T, tcs ...rpcTest) {
...
+	call := func(t *testing.T, store any, tc rpcTest) error {
+		t.Helper()
+		t.Logf("%T.CallContext(ctx, %T, %q, %v...)", s.rpcClient, &tc.want /*i.e. the type*/, tc.method, tc.args)
+		return s.CallContext(ctx, store, tc.method, tc.args...)
+	}
+
 	for _, tc := range tcs {
 		t.Run(tc.method, func(t *testing.T) {
+			if tc.wantErr != "" {
+				err := call(t, nil, tc) // won't unmarshal anything
```
ARR4N: This will panic if the method doesn't actually error and attempts to unmarshal into the `nil`. The test still fails but it becomes hard to debug.

PR #160, opened 2026-02-11
feat: VM restart with database recovery

ARR4N:
Adds support for shutdown and restart without state sync, recovering entirely from the local database.

## Recommended review order

1. Changes to `blocks.Block.Mark{Executed,Synchronous}()` and the new `Block.RestoreExecutionArtefacts()`; plus accompanying tests.
2. Changes to `sae.NewVM()` and `sae.SinceGenesis.Initialize()`.
3. Introduction of `sae/recovery.go` to support (2), and associated test in `recovery_test.go`.
4. All other changes, which are motivated by the above and generally self-contained.

## Mempool rationale

The upstream `legacypool` implementation expects a synchronous blockchain, initially requesting the current block and then updating based on chain-head events, in both cases opening a `state.StateDB` at the latest `types.Header.Root`. In an asynchronous implementation this results in the mempool acting on _settled_, not _executed_ state. So far this has caused two undesirable properties:

1. $\tau$ seconds of empty blocks. Until settled, included transactions remain in the mempool, unblocking `sae.VM.WaitForEvent()`, only to be filtered out by `worstcase`. This also suggests an underlying inefficiency in which every `BuildBlock()` first discards some prefix of already-included transactions.
2. VMs recovered after shutdown may experience a false nonce gap (discovered by `sae.TestRecoverFromDatabase()` in this PR) that don't allow their `BuildBlock()` method to include any transaction from an EOA with included but not settled transactions.

The wrapper returned by `txgossip.NewBlockChain()` addresses this by always serving the latest executed state, regardless of which root is requested. The impossibility of re-orgs makes this safe and efficient (no mempool resets), and addresses (2) entirely. Although this doesn't address all of (1) and some empty blocks and discarded prefixes can occur, it significantly curtails the issue.

### DIFF
--- a/sae/vm_test.go
+++ b/sae/vm_test.go
@@ -553,6 +553,7 @@ func TestEmptyChainConfig(t *testing.T) {
 	for range 5 {
 		sut.runConsensusLoop(t, sut.lastAcceptedBlock(t))
 	}
+	sut.waitUntilExecuted(t, sut.lastAcceptedBlock(t))
 }
 
 func TestSyntacticBlockChecks(t *testing.T) {


blocks/execution.go
```diff
@@ -70,56 +94,104 @@ func (b *Block) MarkExecuted(
...
+// regardless of the upstream trigger. See documentation re ordering invariants
+// for more information.
+//
+// Only the [executionResults] are required. If `setAsHeadBlock` is true then
+// the [ethdb.KeyValueWriter] MUST be true for setting to take effect. If the KV
+// writer also has a `Write()` method (e.g. an [ethdb.Batch]) then it will be
+// called after all `Put()` calls.
+func (b *Block) markExecuted(kv ethdb.KeyValueWriter, e *executionResults, setAsHeadBlock bool, lastExecuted *atomic.Pointer[Block]) error {
+	// Disk
+	if kv != nil {
```
StephenButtolph: imo, rather than handling `kv == nil`, we should separate the Memory updates into a function so that `RestoreExecutionArtefacts` could call that directly.

blocks/settlement.go
```diff
@@ -80,23 +78,29 @@ func (b *Block) markSettled(lastSettled *atomic.Pointer[Block]) error {
...
 	// would also require them to be received as an argument to MarkSynchronous.
-	var rs types.Receipts
-	if err := b.MarkExecuted(db, execTime, time.Time{}, baseFee, rs, ethB.Root(), new(atomic.Pointer[Block])); err != nil {
+	e := &executionResults{
+		byGas: *gastime.New(
+			PreciseTime(hooks, b.Header()),
+			hooks.GasTargetAfter(b.Header()), // target _after_ is a requirement of [Block.MarkExecuted]
+			excessAfter,
+		),
+		receiptRoot:   b.b.ReceiptHash(),
```
alarso16:
```suggestion
		receiptRoot:   ethB.ReceiptHash(),
```

txgossip/blockchain.go
```diff
@@ -42,7 +42,18 @@ func (bc *blockchain) Config() *params.ChainConfig {
...
+	// similar transformation will be necessary for
+	// [blockchain.SubscribeChainHeadEvent] but taking care to update the parent
+	// hash too otherwise the mempool will think that every block is a reorg. An
+	// alternative, which might have its own issues, is to have
+	// [blockchain.StateAt] assume that it receives the settled state, map that
+	// to the executed state at the same block, and open that root instead.
+	last := bc.exec.LastExecuted()
+	h := types.CopyHeader(last.Header())
+	h.Root = last.PostExecutionStateRoot()
+	return h
```
alarso16: This is scary, this will have an updated hash, right? I like the other option you list better

sae/vm.go
```diff
@@ -140,6 +135,25 @@ func NewVM(
...
 		vm.toClose = append(vm.toClose, exec.Close)
+
+		last := executeAfter
+		for b, err := range unexecuted {
+			if err != nil {
+				return nil, err
+			}
+			if err := exec.Enqueue(ctx, b); err != nil {
+				return nil, err
+			}
```
StephenButtolph: Probably fine, but it might be possible for us to hit that warning on enqueue here if the number of blocks to reexecute gets too large

params/params.go
```diff
@@ -21,3 +27,32 @@ const (
...
+
+	copy(key, []byte(rawDBPrefix))
+	copy(key[len(rawDBPrefix):], []byte(namespace))
+
+	key[n-1] = '-'
+	return binary.BigEndian.AppendUint64(key, num)
+}
+
+const (
+	CommitTrieDBEvery     = 1 << commitTrieDBEveryLog2
```
alarso16:
```suggestion
	CommitTrieDBEvery     = 1 << commitTrieDBEveryLog2 // 4096
```

sae/vm.go
```diff
@@ -140,6 +135,25 @@ func NewVM(
...
+			last = b
+		}
+		if err := last.WaitUntilExecuted(ctx); err != nil {
+			return nil, err
+		}
+	}
+
+	// ==========  Pedantry  ==========
+	if err := vm.rebuildBlocksInMemory(lastSynchronous); err != nil {
```
alarso16: We're keeping all blocks in memory since genesis? What's preventing us from just keeping from the last settled?

params/params.go
```diff
@@ -21,3 +27,32 @@ const (
...
+	commitTrieDBEveryLog2 = 12
+	commitTrieDBMask      = CommitTrieDBEvery - 1
+)
+
+func CommitTrieDB(blockNum uint64) bool {
+	return blockNum > 0 && blockNum&commitTrieDBMask == 0
```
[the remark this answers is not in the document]
alarso16: The genesis block I guess is always handled separately, but having one root always on disk sounds nice

blocks/settlement.go
```diff
@@ -80,23 +78,29 @@ func (b *Block) markSettled(lastSettled *atomic.Pointer[Block]) error {
...
+		stateRootPost: ethB.Root(),
+	}
+	if err := e.setBaseFee(ethB.BaseFee()); err != nil {
+		return err
+	}
+	if err := b.markExecuted(db, e, false, nil); err != nil {
```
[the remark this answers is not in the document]
ARR4N: During database recovery it's no longer the head block. That's why `sae.canonicaliseLastSynchronous()` is responsible for setting the synchronous block as head i.f.f. no others have been accepted.

txgossip/blockchain.go
ARR4N: It'll be revisited, hence the verbose TODO. Yup, a different hash, which the mempool will think is a reorg, but only right after construction.

sae/blocks.go
```diff
@@ -349,6 +350,10 @@ func (vm *VM) VerifyBlock(ctx context.Context, bCtx *block.Context, b *blocks.Bl
 	return nil
 }
 
+func (vm *VM) canonicalBlock(num uint64) *types.Block {
+	return rawdb.ReadBlock(vm.db, rawdb.ReadCanonicalHash(vm.db, num), num)
+}
```
StephenButtolph: (See prior comment around method calls on partially initialized values)
```suggestion
func readCanonicalBlock(db ethdb.Reader, num uint64) *types.Block {
	return rawdb.ReadBlock(db, rawdb.ReadCanonicalHash(db, num), num)
}
```

blocks/block.go
```diff
@@ -102,7 +102,8 @@ var (
 	errHashMismatch               = errors.New("block hash mismatch")
 )
 
-func (b *Block) setAncestors(parent, lastSettled *Block) error {
+// SetAncestors sets the block's ancestry while enforcing invariants.
+func (b *Block) SetAncestors(parent, lastSettled *Block) error {
```
alarso16: The comment isn't very helpful (maybe just to satisfy the linter), but why would the parent be nil? Is it just the genesis block?

params/params.go
```diff
@@ -21,3 +27,37 @@ const (
...
+// CommitTrieDB returns whether or not to commit the state trie to disk.
+func CommitTrieDB(blockNum uint64) bool {
+	return blockNum&commitTrieDBMask == 0
+}
+
+// LastCommittedTrieDBHeight returns the largest value <= the argument at which
+// [CommitTrieDB] would have returned true.
+func LastCommittedTrieDBHeight(atOrBefore uint64) uint64 {
+	return atOrBefore &^ commitTrieDBMask
+}
```
StephenButtolph: These feel very out-of-place in the `params` package. Thoughts on making a `saedb` package?

params/params.go
```diff
@@ -21,3 +27,37 @@ const (
...
+const (
+	// CommitTrieDBEvery is the number of blocks between commits of the state
+	// trie to disk.
+	CommitTrieDBEvery     = 1 << commitTrieDBEveryLog2
+	commitTrieDBEveryLog2 = 12
+	commitTrieDBMask      = CommitTrieDBEvery - 1
+)
+
+// CommitTrieDB returns whether or not to commit the state trie to disk.
+func CommitTrieDB(blockNum uint64) bool {
```
StephenButtolph: This name feels odd to me
```suggestion
// ShouldCommitTrieDB returns whether or not to commit the state trie to disk.
func ShouldCommitTrieDB(blockNum uint64) bool {
```

This should make the call site more natural imo
```golang
if num := b.NumberU64(); params.ShouldCommitTrieDB(num) {
	...
	tdb.Commit()
	...
}
```

sae/vm.go
```diff
@@ -140,6 +135,25 @@ func NewVM(
...
+				return nil, err
+			}
+			last = b
+		}
+		if err := last.WaitUntilExecuted(ctx); err != nil {
+			return nil, err
+		}
+	}
+
+	// ==========  Pedantry  ==========
```
StephenButtolph: I feel like `Pedantry` isn't super descriptive here haha

sae/vm.go
alarso16: This warning would certainly be triggered if it wasn't the exact size you put it on startup - maybe set it to the same constant we use to persist with, just so we can't ever mess it up?

And it will be hit on bootstrapping, right? Might want to ignore the warning if we're bootstrapping

sae/recovery.go
```diff
@@ -0,0 +1,148 @@
...
+
+// rebuildBlocksInMemory reinstates internal invariants pertaining to blocks
+// kept in memory. After it returns, all blocks from the
+// [saexec.Executor.LastExecuted] back to the block that it settled will be in
+// memory, with their ancestry populated as required by invariants.
+func (vm *VM) rebuildBlocksInMemory(lastSynchronous *blocks.Block) error {
+	head := vm.exec.LastExecuted()
+	chain := []*blocks.Block{head} // reverse height order
+	blackhole := new(atomic.Pointer[blocks.Block])
+
```
StephenButtolph: Thoughts on adding another helper function here to cleanup the `chain[len(chain)-1]` usage that we use in a couple places?
```golang
last := func() *blocks.Block {
	return chain[len(chain)-1]
}
```
Makes things more clear imo

sae/vm.go
```diff
@@ -84,16 +83,20 @@ type RPCConfig struct {
...
 	lastSynchronous *blocks.Block,
 	sender snowcommon.AppSender,
-) (*VM, error) {
+) (_ *VM, retErr error) {
 	if c.Now == nil {
 		c.Now = time.Now
 	}
+	if err := canonicaliseLastSynchronous(db, lastSynchronous); err != nil {
+		return nil, err
+	}
```
StephenButtolph: Could you be able to explain the rationale on why we `MarkSynchronous` in `SinceGenesis.Initialize` and then canonicalize in `NewVM`?

To me, it feels like it would flow more naturally for `NewVM` to take in a `lastSynchronous *types.Block`, and then inside `NewVM` manage marking the block as synchronous (if needed) and making sure the state is correct.

blocks/invariants.go
```diff
@@ -46,6 +46,10 @@ func (b *Block) WorstCaseBounds() *WorstCaseBounds {
 // execution so no error is returned and execution MUST continue optimistically.
 // Any such log in development will cause tests to fail.
 func (b *Block) CheckBaseFeeBound(actual *uint256.Int) {
+	if b.bounds == nil {
```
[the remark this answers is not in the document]
ARR4N: The block replay at recovery requires execution of all blocks since the last one with an available state root. The `iter.Seq2` returned in `recovery.go` will yield blocks that hit this bit of the code.

txgossip/blockchain.go
ARR4N: Note to reviewers: ignore this file as the changes will be made obsolete by #162, which will be merged before this PR.

sae/recovery.go
```diff
@@ -0,0 +1,148 @@
...
+func (vm *VM) lastBlockWithStateRootAvailable(lastSync *blocks.Block) (*blocks.Block, error) {
+	lastExec := rawdb.ReadHeadHeader(vm.db)
+	num := max(
+		lastSync.NumberU64(),
+		params.LastCommittedTrieDBHeight(lastExec.Number.Uint64()),
+	)
+	if num == lastSync.NumberU64() {
+		return lastSync, nil
+	}
```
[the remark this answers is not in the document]
ARR4N: Can't apply the suggestion due to the earlier refactoring of the methods off of `VM`, but done.

blocks/execution.go
```diff
@@ -70,56 +94,104 @@ func (b *Block) MarkExecuted(
...
+func (b *Block) RestoreExecutionArtefacts(db ethdb.Database) error {
+	buf, err := db.Get(executionResultsKey(b.NumberU64()))
+	if err != nil {
+		return err
+	}
+	e := new(executionResults)
+	if err := e.UnmarshalCanoto(buf); err != nil {
+		return err
+	}
+	e.receipts = rawdb.ReadRawReceipts(db, b.Hash(), b.NumberU64())
```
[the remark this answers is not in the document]
StephenButtolph: `rawdb.ReadReceipts` derives the receipts with the wrong `baseFee`

params/params.go
```diff
@@ -21,3 +27,37 @@ const (
...
+func RawDBKeyForBlock(namespace string, num uint64) []byte {
+	n := len(rawDBPrefix) + len(namespace) + 1 /*hyphen*/
+	key := make([]byte, n, n+8)
+
+	copy(key, []byte(rawDBPrefix))
+	copy(key[len(rawDBPrefix):], []byte(namespace))
+
+	key[n-1] = '-'
+	return binary.BigEndian.AppendUint64(key, num)
+}
```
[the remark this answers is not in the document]
StephenButtolph: > Should we just use prefixdb instead?

Absolutely no.

sae/vm.go
ARR4N: With the introduction of the height-indexed DB the benefit of maintaining invariants became great enough that I've moved it into `NewVM()`. `SinceGenesis.Initialize()` is now extremely simple!

sae/vm.go
```diff
@@ -83,55 +87,70 @@ type RPCConfig struct {
...
+	xdb, err := cfg.Hooks.ExecutionResultsDB(
+		filepath.Join(snowCtx.ChainDataDir, "sae_execution_results"),
+	)
+	if err != nil {
+		return nil, fmt.Errorf("%T.ExecutionResultsDB(%q): %v", cfg.Hooks, snowCtx.ChainDataDir, err)
+	}
+	vm.toClose = append(vm.toClose, xdb.Close)
+
+	{ // ==========  Sync -> Async  ==========
+		if err := lastSynchronous.MarkSynchronous(cfg.Hooks, db, xdb, cfg.ExcessAfterLastSynchronous); err != nil {
```
StephenButtolph: No change required, but it does feel a bit odd for us to always perform these DB writes on startup.

It's probably best for us to just leave it and move on... But I suspect there could be some cleanup here that would (maybe) allow us to further cleanup the `MarkSynchronous` implementation.

I had hoped that we could only call `MarkSynchronous` until we had accepted a following block... Then allowing us to always `SetAsHeadBlock` inside `MarkSynchronous` and then we could unexported  `SetAsHeadBlock`... But it's possible that we need to restart the VM after having accepted (and executed) a block, but not yet having settled a block...

In this case, we wouldn't want to overwrite the head block, but we still need to mark the block as synchronous because it is the last settled block.

All this to say, the shape of the code feels weird to me... But I don't have a better suggestion... So... 🤷

StephenButtolph (commented):
Only one actual reliability concern.

sae/recovery.go
```diff
@@ -0,0 +1,159 @@
...
+		for extended := false; ; extended = true {
+			switch b := lastOf(chain); {
+			case b.Synchronous():
+				return nil
+
+			case b.ExecutedByGasTime().Compare(tm) <= 0:
+				if !extended {
+					return nil
+				}
+				return b.MarkSettled(blackhole)
```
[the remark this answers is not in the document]
ARR4N: > none of the blocks that are settled but without settled called on them are exposed from the results of this function

This is why it worked, but I agree that it was weird. I made a couple of small tweaks for any future reader of the code, one to be more explicit about when not to re-settle (instead of using `!extended`) and the other to maintain the settlement monotonicity.

PR #161, opened 2026-02-12
feat: add gas price estimator 

ceyonur:
Adds the gas price estimator (formerly known as Oracle) to support gas price APIs:
- FeeHistory
- SuggestTipCap

### DIFF
--- a/gasprice/block_cache.go
+++ b/gasprice/block_cache.go
@@ -98,13 +98,10 @@ type blockCache struct {
 }
 
 func newBlockCache(log logging.Logger, backend Backend, size int) *blockCache {
-	// Additional slots in the cache allows processing queries for previous
-	// blocks while new blocks are added concurrently.
-	const extraSlots = 5
 	return &blockCache{
 		log:     log,
 		backend: backend,
-		cache:   lru.NewCache[uint64, *block](size + extraSlots),
+		cache:   lru.NewCache[uint64, *block](size),
 	}
 }
 
--- a/gasprice/estimator.go
+++ b/gasprice/estimator.go
@@ -16,7 +16,6 @@ import (
 	"time"
 
 	"github.com/ava-labs/avalanchego/utils/logging"
-	"github.com/ava-labs/libevm/common"
 	"github.com/ava-labs/libevm/common/math"
 	"github.com/ava-labs/libevm/core"
 	"github.com/ava-labs/libevm/core/types"
@@ -31,7 +30,7 @@ import (
 // Backend that the [Estimator] depends on for chain data.
 type Backend interface {
 	ResolveBlockNumber(bn rpc.BlockNumber) (uint64, error)
-	BlockByNumber(number rpc.BlockNumber) (*types.Block, error)
+	BlockByNumber(bn rpc.BlockNumber) (*types.Block, error)
 	SubscribeChainHeadEvent(ch chan<- core.ChainHeadEvent) event.Subscription
 	LastAcceptedBlock() *blocks.Block
 }
@@ -139,14 +138,11 @@ func NewEstimator(backend Backend, log logging.Logger, c Config) (*Estimator, er
 	// TODO(StephenButtolph): Consider caching upon acceptance rather than execution.
 	events := make(chan core.ChainHeadEvent, 1)
 	sub := backend.SubscribeChainHeadEvent(events)
-	cache := newBlockCache(
-		log,
-		backend,
-		int(max( //nolint:gosec // Overflow would require misconfiguration
-			c.SuggestedTipMaxBlocks,
-			c.HistoryMaxBlocksFromHead+c.HistoryMaxBlocks,
-		)),
-	)
+	// Additional slots in the cache allows processing queries for previous
+	// blocks while new blocks are added concurrently.
+	const extraSlots = 5
+	size := max(c.SuggestedTipMaxBlocks, c.HistoryMaxBlocksFromHead+c.HistoryMaxBlocks) + extraSlots
+	cache := newBlockCache(log, backend, int(size)) //nolint:gosec // Overflow would require misconfiguration
 	go func() {
 		defer sub.Unsubscribe() // `Unsubscribe` can fire twice on Close(), but it's safe to call multiple times.
 		for {
@@ -285,7 +281,7 @@ func (e *Estimator) FeeHistory(
 		last+1,               // Underflow protection for "first" calculation
 	)
 	if blocks == 0 {
-		return common.Big0, nil, nil, nil, nil
+		return big.NewInt(0), nil, nil, nil, nil
 	}
 
 	first := last + 1 - blocks
--- a/gasprice/estimator_test.go
+++ b/gasprice/estimator_test.go
@@ -133,8 +133,7 @@ func (s *SUT) newBlock(tb testing.TB, time uint64, bounds *blocks.WorstCaseBound
 	return blk
 }
 
-func (s *SUT) newTx(tb testing.TB, gas, price uint64) *types.Transaction {
-	tb.Helper()
+func newDynamicFeeTx(gas, price uint64) *types.Transaction {
 	return types.NewTx(&types.DynamicFeeTx{
 		Gas:       gas,
 		GasTipCap: new(big.Int).SetUint64(price),
@@ -276,7 +275,7 @@ func TestSuggestTipCap(t *testing.T) {
 			for _, spec := range test.blocks {
 				txs := make([]*types.Transaction, len(spec.txTips))
 				for i, price := range spec.txTips {
-					txs[i] = sut.newTx(t, 1, price)
+					txs[i] = newDynamicFeeTx(1, price)
 				}
 				sut.newBlock(t, spec.time, nil, txs...)
 			}
@@ -296,11 +295,7 @@ func TestFeeHistory(t *testing.T) {
 		LatestEndTime: gastime.New(time.Now(), 1, math.MaxUint64),
 	}
 	type (
-		txSpec struct {
-			gas   uint64
-			price uint64
-		}
-		blockSpec []txSpec
+		blockSpec []*types.Transaction
 		args      struct {
 			numBlocks   uint64
 			lastBlock   rpc.BlockNumber
@@ -384,7 +379,7 @@ func TestFeeHistory(t *testing.T) {
 			name: "query_genesis",
 			blocks: []blockSpec{
 				{
-					{gas: 21_000, price: nAVAX},
+					newDynamicFeeTx(21_000, nAVAX),
 				},
 			},
 			args: args{
@@ -406,7 +401,7 @@ func TestFeeHistory(t *testing.T) {
 			name: "query_latest",
 			blocks: []blockSpec{
 				{
-					{gas: 21_000, price: nAVAX},
+					newDynamicFeeTx(21_000, nAVAX),
 				},
 			},
 			args: args{
@@ -428,10 +423,10 @@ func TestFeeHistory(t *testing.T) {
 			name: "query_too_old_block",
 			blocks: []blockSpec{
 				{
-					{gas: 21_000, price: nAVAX},
+					newDynamicFeeTx(21_000, nAVAX),
 				},
 				{
-					{gas: 100_000, price: nAVAX},
+					newDynamicFeeTx(100_000, nAVAX),
 				},
 			},
 			args: args{
@@ -446,14 +441,14 @@ func TestFeeHistory(t *testing.T) {
 			name: "query_max_blocks_with_percentiles",
 			blocks: []blockSpec{
 				{
-					{gas: 21_000, price: nAVAX},
+					newDynamicFeeTx(21_000, nAVAX),
 				},
 				{
-					{gas: 100_000, price: nAVAX},
-					{gas: 100_000, price: 2 * nAVAX},
-					{gas: 100_000, price: 3 * nAVAX},
-					{gas: 100_000, price: 4 * nAVAX},
-					{gas: 100_000, price: 5 * nAVAX},
+					newDynamicFeeTx(100_000, nAVAX),
+					newDynamicFeeTx(100_000, 2*nAVAX),
+					newDynamicFeeTx(100_000, 3*nAVAX),
+					newDynamicFeeTx(100_000, 4*nAVAX),
+					newDynamicFeeTx(100_000, 5*nAVAX),
 				},
 			},
 			args: args{
@@ -483,11 +478,7 @@ func TestFeeHistory(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			sut := newSUT(t, cfg)
 			for _, txSpecs := range tt.blocks {
-				txs := make([]*types.Transaction, len(txSpecs))
-				for i, tx := range txSpecs {
-					txs[i] = sut.newTx(t, tx.gas, tx.price)
-				}
-				sut.newBlock(t, 0, bounds, txs...)
+				sut.newBlock(t, 0, bounds, txSpecs...)
 			}
 
 			a := tt.args
--- a/worstcase/state_test.go
+++ b/worstcase/state_test.go
@@ -109,7 +109,7 @@ func TestMultipleBlocks(t *testing.T) {
 
 	state := sut.State
 	lastHash := sut.genesis.Hash()
-	wantNextGasTime := sut.genesis.ExecutedByGasTime().Clone()
+	wantLatestEndTime := sut.genesis.ExecutedByGasTime().Clone()
 
 	const importedAmount = 10
 	type op struct {
@@ -283,7 +283,7 @@ func TestMultipleBlocks(t *testing.T) {
 			Time:       block.time,
 		}
 
-		wantNextGasTime.BeforeBlock(sut.hooks, header)
+		wantLatestEndTime.BeforeBlock(sut.hooks, header)
 		require.NoErrorf(t, state.StartBlock(header), "StartBlock(%d)", i)
 		require.Equalf(t, block.wantBaseFee, state.BaseFee(), "base fee after StartBlock(%d)", i)
 		require.Equalf(t, block.wantGasLimit, state.GasLimit(), "gas limit after StartBlock(%d)", i)
@@ -298,11 +298,11 @@ func TestMultipleBlocks(t *testing.T) {
 
 		got, err := state.FinishBlock()
 		require.NoError(t, err, "FinishBlock()")
-		require.NoError(t, wantNextGasTime.AfterBlock(gas.Gas(state.GasUsed()), sut.hooks, header), "AfterBlock()")
+		require.NoError(t, wantLatestEndTime.AfterBlock(gas.Gas(state.GasUsed()), sut.hooks, header), "AfterBlock()")
 
 		want := &blocks.WorstCaseBounds{
 			MaxBaseFee:    block.wantBaseFee,
-			LatestEndTime: wantNextGasTime.Clone(),
+			LatestEndTime: wantLatestEndTime.Clone(),
 		}
 		for _, bals := range block.wantMinSenderBalances {
 			uBals := make(map[common.Address]*uint256.Int)


gasprice/gasprice.go
```diff
@@ -0,0 +1,171 @@
...
+	"github.com/ava-labs/libevm/rpc"
+	"golang.org/x/exp/slices"
+)
+
+// OracleBackend includes all necessary background APIs for oracle.
+type OracleBackend interface {
+	ResolveBlockNumber(bn rpc.BlockNumber) (uint64, error)
+	HeaderByNumber(ctx context.Context, number rpc.BlockNumber) (*types.Header, error)
+	BlockByNumber(ctx context.Context, number rpc.BlockNumber) (*types.Block, error)
+	SubscribeChainHeadEvent(ch chan<- core.ChainHeadEvent) event.Subscription
```
StephenButtolph: I don't think we need `SubscribeChainHeadEvent`. We never have reorgs and it is only used for reorg detection.

gasprice/gasprice.go
```diff
@@ -0,0 +1,171 @@
...
+		slices.SortFunc(tipResults, func(a, b *big.Int) int { return a.Cmp(b) })
+		price = tipResults[(len(tipResults)-1)*oracle.cfg.Percentile/100]
+	}
+
+	if price.Cmp(oracle.cfg.MaxPrice) > 0 {
+		price = new(big.Int).Set(oracle.cfg.MaxPrice)
+	}
+	if price.Cmp(oracle.cfg.MinPrice) < 0 {
+		price = new(big.Int).Set(oracle.cfg.MinPrice)
+	}
```
StephenButtolph: I think we can use `math.BigMin` and `math.BigMax` here. They don't clone the values, but I also don't think there is any reason we clone the values here anyways.

gasprice/config.go
```diff
@@ -0,0 +1,151 @@
...
+	MaxLookbackSeconds uint64
+	// MaxCallBlockHistory specifies the maximum number of blocks that can be
+	// fetched in a single eth_feeHistory call.
+	MaxCallBlockHistory uint64
+	// MaxBlockHistory specifies the furthest back behind the last accepted block that can
+	// be requested by fee history.
+	MaxBlockHistory uint64
+	MaxPrice        *big.Int `toml:",omitempty"`
+	MinPrice        *big.Int `toml:",omitempty"`
+	MinGasUsed      *big.Int `toml:",omitempty"`
```
StephenButtolph: Why do these have `toml` references?

StephenButtolph: `MinGasUsed` is unused

gasprice/feehistory.go
```diff
@@ -0,0 +1,219 @@
...
+	}
+	lastBlock, blocks, err := oracle.resolveBlockRange(unresolvedLastBlock, blocks)
+	if err != nil || blocks == 0 {
+		return common.Big0, nil, nil, nil, err
+	}
+	oldestBlock := lastBlock + 1 - blocks
+
+	var (
+		reward       = make([][]*big.Int, blocks)
+		baseFee      = make([]*big.Int, blocks)
```
StephenButtolph: Are we deviating from the expected behavior here by not providing the next base fee? I know that we can't know it for sure, but we could return an upper-bound on the next block's base fee.

gasprice/fee_info_provider.go
```diff
@@ -0,0 +1,111 @@
...
+
+// additional slots in the header cache to allow processing queries
+// for previous blocks (with full number of blocks desired) if new
+// blocks are added concurrently.
+const feeCacheExtraSlots = 5
+
+type feeInfoProvider struct {
+	cache          *lru.Cache
+	backend        OracleBackend
+	newHeaderAdded func() // callback used in tests
```
StephenButtolph: I don't think we should be adding in code specific to testing in prod, we should figure out how to test our code through the exposed interface imo

gasprice/config.go
```diff
@@ -0,0 +1,151 @@
...
+	Percentile int
+	// MaxLookbackSeconds specifies the maximum number of seconds that current timestamp
+	// can differ from block timestamp in order to be included in gas price estimation
+	MaxLookbackSeconds uint64
+	// MaxCallBlockHistory specifies the maximum number of blocks that can be
+	// fetched in a single eth_feeHistory call.
+	MaxCallBlockHistory uint64
+	// MaxBlockHistory specifies the furthest back behind the last accepted block that can
+	// be requested by fee history.
+	MaxBlockHistory uint64
```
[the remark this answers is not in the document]
ceyonur: Wym? This parameter does restrict querying older blocks in https://github.com/ava-labs/strevm/blob/ceyonur/gasprice-oracle/gasprice/feehistory.go#L113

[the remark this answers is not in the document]
ceyonur: If you're asking why do we have different mechanism for restriction:

1- MaxCallBlockHistory: "how many blocks can you fetch in one call"
2- MaxBlockHistory: "how far back from the head can you ever look"

``` 	
// DefaultMaxBlockHistory is chosen to be a value larger than the required
// fee lookback window that MetaMask uses (20k blocks).
```
I always thought this chosen big enough so we can also limit how far back we fetch.

gasprice/feehistory.go
ceyonur: I think we have already deviated in Coreth. but probably makes sense to recalculate it now. Although I don't know how this would work with acp-224 dynamic parameters.

gasprice/fee_info_provider.go
```diff
@@ -0,0 +1,178 @@
...
+// block and returns the value as a feeInfo struct.
+// It adds the entry to the cache if missing.
+func (f *feeInfoProvider) getFeeInfo(ctx context.Context, number uint64) (*feeInfo, error) {
+	feeInfo, ok := f.cache.Get(number)
+	if ok {
+		return feeInfo, nil
+	}
+
+	// on cache miss, read from database
+	block, err := f.backend.BlockByNumber(ctx, rpc.BlockNumber(number)) //nolint:gosec // block numbers are always within int64 range
```
JonathanOppenheimer: Here, `getFeeInfo` assumes `BlockByNumber` always returns a non-nil block and immediately calls f.addBlock(block), but this isn't a valid assumption -- the backend is `(nil, nil)` for "not found" cases, so a missing block in the requested range will cause a nil-pointer panic, when I think we can instead have a clean not found response.

gasprice/config.go
```diff
@@ -0,0 +1,137 @@
...
+	// a single call to eth_feeHistory.
+	DefaultMaxCallBlockHistory = 2048
+	// DefaultMaxBlockHistory is the number of blocks from the last accepted
+	// block that can be fetched in eth_feeHistory.
+	//
+	// DefaultMaxBlockHistory is chosen to be a value larger than the required
+	// fee lookback window that MetaMask uses (20k blocks).
+	DefaultMaxBlockHistory = 25_000
+	// FeeHistoryCacheSize is chosen to be some value larger than
+	// [DefaultMaxBlockHistory] to ensure all block lookups can be cached when
```
JonathanOppenheimer: (nit) You inconsistently use the `[]` -- I would check these are correct and be consistent with it

gasprice/config.go
```diff
@@ -0,0 +1,137 @@
...
+	if maxBlockHistory < 1 {
+		return nil, fmt.Errorf("max block history (%d) is less than 1", maxBlockHistory)
+	}
+	return options.Func[config](func(c *config) {
+		c.MaxBlockHistory = maxBlockHistory
+	}), nil
+}
+
+// WithMaxPrice sets the maximum suggested tip.
+func WithMaxPrice(maxPrice *big.Int) (EstimatorOption, error) {
```
JonathanOppenheimer: Should `min <= max` be enforced?

gasprice/config.go
```diff
@@ -0,0 +1,137 @@
...
+	if maxPrice == nil || maxPrice.Sign() <= 0 {
+		return nil, fmt.Errorf("max price (%v) is nil or non-positive", maxPrice)
+	}
+	return options.Func[config](func(c *config) {
+		c.MaxPrice = maxPrice
+	}), nil
+}
+
+// WithMinPrice sets the minimum suggested tip.
+func WithMinPrice(minPrice *big.Int) (EstimatorOption, error) {
```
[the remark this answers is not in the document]
ceyonur: https://github.com/ava-labs/strevm/pull/161#discussion_r2816747845

gasprice/estimator.go
```diff
@@ -0,0 +1,303 @@
...
+
+	last, blocks, err := e.clampBlockRange(ctx, unresolvedLastBlock, blocks)
+	if err != nil || blocks == 0 {
+		return common.Big0, nil, nil, nil, err
+	}
+	first := last + 1 - blocks
+
+	var (
+		reward       = make([][]*big.Int, blocks)
+		baseFee      = make([]*big.Int, blocks)
```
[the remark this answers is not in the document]
ceyonur: there is a discussion here https://github.com/ava-labs/strevm/pull/161#discussion_r2816754232

gasprice/estimator.go
```diff
@@ -0,0 +1,303 @@
...
+	}
+	// if we could not find the feeInfo in the history cache, we check the feeInfoProvider's cache.
+	// as it might be found there if it's a recent block.
+	// Note: do not fetch the block with feeInfoProvider.getFeeInfo because it can evict recent blocks from the history cache.
+	feeInfo, ok := e.feeInfoProvider.cache.Get(number)
+	if ok {
+		e.historyCache.Put(number, feeInfo)
+		return feeInfo, nil
+	}
+	block, err := e.backend.BlockByNumber(ctx, rpc.BlockNumber(number)) //nolint:gosec // block numbers are always within int64 range
```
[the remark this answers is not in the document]
ceyonur: hmm nice catch, I thought we always return `errFutureBlockNotResolved` from that backend.

gasprice/estimator_test.go
```diff
@@ -0,0 +1,462 @@
...
+		expErr       error
+	}{
+		// Standard libevm tests
+		{maxCallBlock: 0, maxBlock: 1000, count: 10, last: 30, percent: nil, expFirst: 21, expCount: 10, expErr: nil},
+		{maxCallBlock: 0, maxBlock: 1000, count: 10, last: 30, percent: []float64{0, 10}, expFirst: 21, expCount: 10, expErr: nil},
+		{maxCallBlock: 0, maxBlock: 1000, count: 10, last: 30, percent: []float64{20, 10}, expFirst: 0, expCount: 0, expErr: errBadPercentile},
+		{maxCallBlock: 0, maxBlock: 1000, count: 1000000000, last: 30, percent: nil, expFirst: 0, expCount: 31, expErr: nil},
+		{maxCallBlock: 0, maxBlock: 1000, count: 1000000000, last: rpc.PendingBlockNumber, percent: nil, expFirst: 0, expCount: 33, expErr: nil},
+		{maxCallBlock: 0, maxBlock: 1000, count: 10, last: 40, percent: nil, expFirst: 0, expCount: 0, expErr: errFutureBlock},
+		{maxCallBlock: 0, maxBlock: 1000, count: 10, last: 40, percent: nil, expFirst: 0, expCount: 0, expErr: errFutureBlock},
```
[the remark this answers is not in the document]
ceyonur: the upstream test cases included a `pending` flag which is removed here, made these cases identical.

gasprice/estimator_test.go
```diff
@@ -0,0 +1,462 @@
...
+		{maxCallBlock: 0, maxBlock: 2, count: 100, last: rpc.PendingBlockNumber, percent: []float64{0, 10}, expFirst: 31, expCount: 2, expErr: nil},
+		{maxCallBlock: 0, maxBlock: 2, count: 100, last: 32, percent: []float64{0, 10}, expFirst: 31, expCount: 2, expErr: nil},
+		// In SAE backend, `pending` resolves to accepted head (not head+1).
+		{maxCallBlock: 0, maxBlock: 1000, count: 1, last: rpc.PendingBlockNumber, percent: nil, expFirst: 32, expCount: 1, expErr: nil},
+		// With count=2, pending spans [head-1, head].
+		{maxCallBlock: 0, maxBlock: 1000, count: 2, last: rpc.PendingBlockNumber, percent: nil, expFirst: 31, expCount: 2, expErr: nil},
+		// Same behavior when "pending" mode is enabled in test matrix.
+		{maxCallBlock: 0, maxBlock: 1000, count: 2, last: rpc.PendingBlockNumber, percent: nil, expFirst: 31, expCount: 2, expErr: nil},
```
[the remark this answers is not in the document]
ceyonur: because "pending" is removed and left these cases identical. thanks.

gasprice/feehistory.go
```diff
@@ -0,0 +1,219 @@
...
+				return common.Big0, nil, nil, nil, err
+			}
+			// getting no block and no error means we are requesting into the future
+			if block == nil {
+				if i == 0 {
+					return common.Big0, nil, nil, nil, nil
+				}
+				firstMissing = i
+				break
+			}
```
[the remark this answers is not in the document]
ceyonur: I ended up restoring this as it appears the backend can actually return nil blocks if they're not found. 
https://github.com/ava-labs/strevm/pull/161#discussion_r2816983854

gasprice/estimator.go
```diff
@@ -0,0 +1,342 @@
...
+	if c.MaxSuggestedTip == nil {
+		return errNilMaxSuggestedTip
+	}
+	if c.SuggestedTipPercentile <= 0 || c.SuggestedTipPercentile > 1 {
+		return errBadTipPercentile
+	}
+	if c.MinSuggestedTip.Cmp(c.MaxSuggestedTip) > 0 {
+		return errMinTipExceedsMax
+	}
+	return nil
```
StephenButtolph: I feel like this would be a bit cleaner as a switch.

gasprice/estimator_test.go
```diff
@@ -0,0 +1,561 @@
...
+	nowSec := uint64(clk.Unix()) //nolint:gosec // Guaranteed to be positive
+
+	newBlock := func(sut *SUT, time uint64, txs ...*types.Transaction) *blocks.Block {
+		t.Helper()
+		return sut.newBlock(t, time, nil, txs...)
+	}
+	newTx := func(sut *SUT, price uint64) *types.Transaction {
+		t.Helper()
+		return sut.newTx(t, 1, price)
+	}
```
StephenButtolph: These functions reference the wrong `t` (they are run in the subtest but use the main test's `t`)

gasprice/estimator.go
```diff
@@ -0,0 +1,342 @@
...
+		c:          c,
+		lastPrice:  c.MinSuggestedTip,
+		sub:        sub,
+		blockCache: cache,
+	}, nil
+}
+
+// SuggestTipCap recommends a priority-fee (tip) for new transactions based on
+// tips from recently accepted transactions.
+func (e *Estimator) SuggestTipCap(ctx context.Context) (tip *big.Int, _ error) {
```
StephenButtolph: Should we update the name of this function now to `SuggestGasTipCap` to align with the backend expectations?

gasprice/estimator.go
```diff
@@ -0,0 +1,342 @@
...
+		}
+		b := e.blockCache.getBlock(ctx, n)
+		if b == nil || b.timestamp < recentUnix {
+			break
+		}
+		tips = append(tips, b.txs...)
+	}
+
+	price := lastPrice
+	if n := len(tips); n > 0 {
```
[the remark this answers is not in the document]
ceyonur: I think both sides have valid arguments:
1- Keeping it lastPrice is more on side of increasing likelyhood of being included after a cycle of empty blocks or no blocks. 
2- using `MinSuggestedTip` is more optimistic that implies "if there was no txs/blocks then there is not a demand". 

I think I'd prefer to increase the likelyhood even it comes with a increased tip in those rare cases. (this is also what upstream geth used/advocates.)

I added the case `no_transactions_fallback_to_last_price`.

gasprice/estimator_test.go
```diff
@@ -0,0 +1,561 @@
...
+	type results struct {
+		height      *big.Int
+		rewards     [][]*big.Int
+		baseFees    []*big.Int
+		portionFull []float64
+		err         error
+	}
+	tests := []struct {
+		name   string
+		blocks func(*SUT) []*blocks.Block
```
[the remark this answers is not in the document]
ceyonur: IMO stateful tests are easier/cleaner to implement but hard to follow. I'd prefer stateless tests as they're easier to debug and add new cases. I also think table-like cases should be stateless otherwise it's defeating the purpose. 

I've refactored tests so they no longer use test functions for `SUT` and `t` but rather define minimal specs for blocks and txs.

gasprice/block_cache.go
```diff
@@ -0,0 +1,137 @@
...
+	var (
+		txIndex = 0
+		sumGas  = b.txs[0].gas
+	)
+	for i, p := range percentiles {
+		threshold := uint64(float64(b.gasUsed) * p / 100)
+		for sumGas < threshold && txIndex < len(b.txs)-1 {
+			txIndex++
+			sumGas += b.txs[txIndex].gas
+		}
```
StephenButtolph: I doubt this will actually happen, but we are being fairly performance conscious in this part of the code... So we could add this:
```suggestion
		// TODO:(StephenButtolph): Improve from `O(txs + percentiles)` to
		// `O(percentiles * log(txs))` by binary searching for each threshold if
		// networks with large blocks encounter performance degradation.
		for sumGas < threshold && txIndex < len(b.txs)-1 {
			txIndex++
			sumGas += b.txs[txIndex].gas
		}
```

gasprice/estimator_test.go
```diff
@@ -0,0 +1,601 @@
...
+		NextGasTime: gastime.New(time.Now(), 1, math.MaxUint64),
+	}
+	type txSpec struct {
+		gas   uint64
+		price uint64
+	}
+
+	type blockSpec struct {
+		txs []txSpec
+	}
```
StephenButtolph: I think we can reduce at least a bit of the verbosity by removing the struct creation... But it'll also make us specify `txSpec` more... So again feel free to reject the suggestion
```suggestion
	type blockSpec []txSpec
```

StephenButtolph (commented):
My comments are getting pretty "nit"-y.

I'll defer to @ARR4N on the final say of the tests. The core logic lgtm

gasprice/estimator.go
```diff
@@ -0,0 +1,337 @@
...
+		return fmt.Errorf("%w: requested %d percentiles, max %d", errBadPercentile, len(percentiles), maxPercentiles)
+	}
+	for i, p := range percentiles {
+		if p < 0 || p > 100 {
+			return fmt.Errorf("%w: value %f at index %d", errBadPercentile, p, i)
+		}
+		if i > 0 && p <= percentiles[i-1] {
+			return fmt.Errorf("%w: index %d (%f) must be greater than index %d (%f)", errBadPercentile, i, p, i-1, percentiles[i-1])
+		}
+	}
```
ARR4N:
```suggestion
	if !slices.IsSorted(percentiles) {
		return fmt.Errorf("percentiles %v not sorted", percentiles)
	}
	for i, p := range percentiles {
		if p < 0 || p > 100 {
			return fmt.Errorf("%w: value %f at index %d", errBadPercentile, p, i)
		}
	}
```

Using the existing, named function is much clearer.

gasprice/estimator.go
```diff
@@ -0,0 +1,337 @@
...
+	// SuggestedTipPercentile, in the range (0, 1], specifies what percentile of
+	// recent tips.
+	SuggestedTipPercentile float64
+	MaxSuggestedTip        *big.Int
+
+	// SuggestedTipMaxBlocks specifies the maximum number of recent blocks to fetch
+	// for [Estimator.SuggestTipCap].
+	SuggestedTipMaxBlocks uint64
+	// SuggestedTipMaxDuration specifies how long a block is considered recent
+	// for [Estimator.SuggestTipCap].
```
ARR4N:
```suggestion
	// for [Estimator.SuggestGasTipCap].
	SuggestedTipMaxBlocks uint64
	// SuggestedTipMaxDuration specifies how long a block is considered recent
	// for [Estimator.SuggestGasTipCap].
```

gasprice/estimator.go
```diff
@@ -0,0 +1,337 @@
...
+			e.c.HistoryMaxBlocksFromTip,
+		)
+	}
+	blocks = min(
+		blocks,               // requested value
+		e.c.HistoryMaxBlocks, // DoS protection
+		last+1,               // Underflow protection for "first" calculation
+	)
+	if blocks == 0 {
+		return common.Big0, nil, nil, nil, nil
```
ARR4N:
```suggestion
		return big.NewInt(0), nil, nil, nil, nil
```

gasprice/estimator.go
```diff
@@ -0,0 +1,337 @@
...
+	ResolveBlockNumber(bn rpc.BlockNumber) (uint64, error)
+	BlockByNumber(ctx context.Context, number rpc.BlockNumber) (*types.Block, error)
+	SubscribeChainHeadEvent(ch chan<- core.ChainHeadEvent) event.Subscription
+	LastAcceptedBlock() *blocks.Block
+}
+
+// Config allows parameterizing an [Estimator].
+type Config struct {
+	// Now returns the current time. If nil, defaults to [time.Now]
+	Now func() time.Time
```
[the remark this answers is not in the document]
StephenButtolph: > It's only used for tests so a consumer of this package has no need for it to be exposed.

Wouldn't it make sense for `NewVM` to specify this so that we don't have multiple time sources during VM testing?

gasprice/estimator.go
```diff
@@ -0,0 +1,337 @@
...
+	case c.MaxSuggestedTip == nil:
+		return errNilMaxSuggestedTip
+	case c.SuggestedTipPercentile < 0 || c.SuggestedTipPercentile > 1:
+		return errBadTipPercentile
+	case c.MinSuggestedTip.Cmp(c.MaxSuggestedTip) > 0:
+		return errMinTipExceedsMax
+	}
+	return nil
+}
+
```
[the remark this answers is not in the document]
StephenButtolph: > Why is that?

Because `ethereum.FeeHistoryReader` and `ethereum.GasPricer1559` are client interfaces, not server interfaces. `Estimator` implements a subset of the `ethapi.Backend` interface.

gasprice/estimator.go
```diff
@@ -0,0 +1,337 @@
...
+//   - The first block of the actually processed range.
+//   - The tips paid for each percentile of the cumulative gas limits of the
+//     transactions in each block.
+//   - The baseFee of each block.
+//   - The portion that each block was filled.
+func (e *Estimator) FeeHistory(
+	ctx context.Context,
+	blocks uint64,
+	unresolvedLastBlock rpc.BlockNumber,
+	rewardPercentiles []float64,
```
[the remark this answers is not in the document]
StephenButtolph: Yeah we can't change the `FeeHistory` expectations... Although we _could_ change the config expectations. Should we just document the rationale? Or would it make more sense to change the config so that the implementations would be aligned?

gasprice/estimator.go
```diff
@@ -0,0 +1,337 @@
...
+// DefaultConfig returns a [Config] with all fields set to their default values.
+func DefaultConfig() Config {
+	return Config{
+		Now:                     time.Now,
+		MinSuggestedTip:         big.NewInt(1 * params.Wei),
+		SuggestedTipPercentile:  .4,
+		MaxSuggestedTip:         big.NewInt(150 * params.Wei),
+		SuggestedTipMaxBlocks:   20,
+		SuggestedTipMaxDuration: time.Minute,
+		// Chosen to be larger than the fee lookback window that MetaMask uses (20k blocks).
```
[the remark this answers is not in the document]
StephenButtolph: > Why 0.4?

This should be documented 👍. 

Some people have previously been concerned that users could cause a self-induced fee spiral. So this number is set to less than 50%. It is set to 60% in geth. So the 10% _over_ 50% was turned into 10% _under_ 50%.

In practice, it shouldn't matter much either way... But it's easier to set an arbitrary number to something people are happy with rather than setting an arbitrary number to something people aren't happy with.

> why longer than the MetaMask window?

So that metamask works, as they depend on this API... Is there anything actionable to change to the comment to make this more clear?

gasprice/estimator.go
```diff
@@ -0,0 +1,337 @@
...
+	cache := newBlockCache(
+		log,
+		backend,
+		int(max( //nolint:gosec // Overflow would require misconfiguration
+			c.SuggestedTipMaxBlocks,
+			c.HistoryMaxBlocksFromTip+c.HistoryMaxBlocks,
+		)),
+	)
+	go func() {
+		defer sub.Unsubscribe() // This can fire twice on Close(), but it's fine.
```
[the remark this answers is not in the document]
StephenButtolph: Perhaps:
```suggestion
		defer sub.Unsubscribe() // Unsubscribe is safe to call multiple times.
```

gasprice/estimator.go
```diff
@@ -0,0 +1,337 @@
...
+		gasUsedRatio = append(gasUsedRatio, float64(b.gasUsed)/float64(b.gasLimit))
+	}
+	if last == head {
+		bounds := headBlock.WorstCaseBounds()
+		if bounds == nil {
+			return nil, nil, nil, nil, errMissingWorstCaseBounds
+		}
+		baseFee = append(baseFee, bounds.NextGasTime.BaseFee().ToBig())
+	} else if b := e.blockCache.getBlock(ctx, last+1); b != nil {
+		baseFee = append(baseFee, b.baseFee)
```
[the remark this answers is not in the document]
StephenButtolph: In ethereum this makes more sense than it does for us.

Ethereum knows the next block's base fee (from the EIP-1559 formula). So geth includes it here.

We don't know what the next block's base fee is, but we don't want to deviate from geth's behavior of adding one more baseFee than the other slices. So we return the best estimate we can give.

blocks/blockstest/chain.go
```diff
@@ -8,23 +8,32 @@
...
 	"github.com/ava-labs/libevm/libevm/options"
 	"github.com/ava-labs/libevm/params"
+	"github.com/ava-labs/libevm/rpc"
 
 	"github.com/ava-labs/strevm/blocks"
 )
 
 // A ChainBuilder builds a chain of blocks, maintaining necessary invariants.
+//
+// It is not safe for concurrent use.
```
StephenButtolph: I had originally though this would cause issues with the execution cache warmer in `NewEstimator`, but it isn't an issue because we don't actually use the backend in the goroutine (we just take the block from the channel).

So this lgtm 👍.

gasprice/estimator.go
ARR4N: I dunno if I want to laugh or to cry 🤣 🤦

gasprice/block_cache.go
```diff
@@ -0,0 +1,137 @@
...
+// getBlock returns the block at height n. If the block does not exist, it will
+// return nil.
+func (b *blockCache) getBlock(ctx context.Context, n uint64) *block {
+	if blk, ok := b.cache.Get(n); ok {
+		return blk
+	}
+
+	blk, err := b.backend.BlockByNumber(context.WithoutCancel(ctx), rpc.BlockNumber(n)) //nolint:gosec // block numbers were previously resolved
+	if err != nil {
+		b.log.Error("fetching BlockByNumber",
```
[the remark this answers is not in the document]
ARR4N: @JonathanOppenheimer every warning should be considered an error in tests and in compilation (i.e. well before production). When actually running the code then it's ok to be a warning.

gasprice/block_cache.go
```diff
@@ -0,0 +1,140 @@
...
+}
+
+// getBlock returns the block at height n. If the block does not exist, it will
+// return nil.
+func (b *blockCache) getBlock(ctx context.Context, n uint64) *block {
+	if blk, ok := b.cache.Get(n); ok {
+		return blk
+	}
+
+	blk, err := b.backend.BlockByNumber(context.WithoutCancel(ctx), rpc.BlockNumber(n)) //nolint:gosec // block numbers were previously resolved
```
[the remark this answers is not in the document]
ARR4N: This package (and PR) defines the `Backend` that it requires, and it's possible to achieve a `BlockByNumber(rpc.BlockNumber)` method (the `Context` parameter is only there because `geth` requires it, but it's never actually used). Yes, it will require a little bit of plumbing to munge the method name to make `*ethAPIBackend` compliant, but it will make this part of the code much clearer.

Something like:

```go
func (b *ethAPIBackend) BlockByNumber(ctx context.Context, n rpc.BlockNumber) (*types.Block, error) {
	return b.BlockByNumberNoCtx(n)
}

func (b *ethAPIBackend) BlockByNumberNoCtx(n rpc.BlockNumber) (*types.Block, error) {
	return readByNumber(b, n, neverErrs(rawdb.ReadBlock))
}
```

That said, the double method naming is gross. It's probably better achieved with an embedded type that implements without the context and a wrapping method that calls it.

gasprice/estimator_test.go
```diff
@@ -0,0 +1,517 @@
...
+	tb.Helper()
+
+	db := rawdb.NewMemoryDatabase()
+	xdb := saetest.NewExecutionResultsDB()
+	config := saetest.ChainConfig()
+	signer := types.LatestSigner(config)
+	wallet := saetest.NewUNSAFEWallet(tb, 1, signer)
+	alloc := saetest.MaxAllocFor(wallet.Addresses()...)
+	genesis := blockstest.NewGenesis(tb, db, xdb, config, alloc)
+	builder := blockstest.NewChainBuilder(config, genesis)
```
ARR4N: (super-nit, optional) The rest of the code typically calls this variable `chain`.

gasprice/estimator_test.go
```diff
@@ -0,0 +1,517 @@
...
+func newSUT(tb testing.TB, c Config) *SUT {
+	tb.Helper()
+
+	db := rawdb.NewMemoryDatabase()
+	xdb := saetest.NewExecutionResultsDB()
+	config := saetest.ChainConfig()
+	signer := types.LatestSigner(config)
+	wallet := saetest.NewUNSAFEWallet(tb, 1, signer)
+	alloc := saetest.MaxAllocFor(wallet.Addresses()...)
+	genesis := blockstest.NewGenesis(tb, db, xdb, config, alloc)
```
ARR4N:
```suggestion
	genesis := blockstest.NewGenesis(tb, db, xdb, config, types.GenesisAlloc{})
```

The sender of the transaction is irrelevant so `SUT.newTx()` can just return `types.NewTransaction(&types.DynamicFeeTx{...})` and the signer, wallet, and allocation are just noise.

gasprice/estimator_test.go
```diff
@@ -0,0 +1,517 @@
...
+	avax  = params.Ether
+	nAVAX = params.GWei
+	aAVAX = params.Wei
+)
+
+func TestSuggestTipCap(t *testing.T) {
+	cfg := DefaultConfig()
+	cfg.MinSuggestedTip = big.NewInt(aAVAX)
+	cfg.MaxSuggestedTip = big.NewInt(avax)
+	clk := time.Unix(100, 0)
```
ARR4N: 🐔 _bagaaak_

gasprice/estimator_test.go
```diff
@@ -0,0 +1,517 @@
...
+				{
+					time:   nowSec - 10,
+					txTips: []uint64{3 * nAVAX, 2 * nAVAX},
+				},
+				{
+					time:   nowSec,
+					txTips: []uint64{4 * nAVAX},
+				},
+			},
+			want: big.NewInt(2 * nAVAX),
```
ARR4N: Just confirming that I've fully grokked this: because it's the 40th percentile?

gasprice/estimator_test.go
```diff
@@ -0,0 +1,517 @@
...
+				{
+					time:   nowSec - (uint64(cfg.SuggestedTipMaxDuration.Seconds()) + 1),
+					txTips: []uint64{math.MaxUint64, math.MaxUint64, math.MaxUint64},
+				},
+				{
+					time:   nowSec,
+					txTips: []uint64{nAVAX},
+				},
+			},
+			want: big.NewInt(1 * nAVAX),
```
ARR4N:
```suggestion
			want: big.NewInt(nAVAX),
```

gasprice/estimator_test.go
ceyonur: yes [1,2,3,4] > 2

PR #168, opened 2026-02-17
feat: state-based RPC methods (e.g. `eth_call`)

ARR4N:
Introduces `GetEVM()` and `StateAndHeaderByNumber{,OrHash}()` methods on `*sae.ethAPIBackend`, enabling state-based methods like `eth_call` (used for testing). The latter necessitated the more general-purpose `resolveBlockNumberOrHash()`, which has its own unit test because exercising all paths via the API is too cumbersome.

### DIFF
--- a/sae/rpc_test.go
+++ b/sae/rpc_test.go
@@ -1045,7 +1045,7 @@ func TestResolveBlockNumberOrHash(t *testing.T) {
 			be := sut.rawVM.apiBackend
 			gotNum, gotHash, err := be.resolveBlockNumberOrHash(tt.nOrH)
 			t.Logf("%T.resolveBlockNumberOrhash(%+v)", be, tt.nOrH) // avoids having to repeat in failure messages
-			require.NoError(t, err)
+			require.ErrorIs(t, err, tt.wantErr)
 			assert.Equal(t, tt.wantNum, gotNum)
 			assert.Equal(t, tt.wantHash, gotHash)
 		})


ARR4N:
The CI error seen on 386820cd5f08d974e0286eec1b16cff93dc2f470 is a known flake, fixed by #175.

sae/rpc_stateful.go
```diff
@@ -0,0 +1,83 @@
...
+
+type coinbaseAsAuthor struct {
+	consensus.Engine
+}
+
+func (*coinbaseAsAuthor) Author(h *types.Header) (common.Address, error) {
+	return h.Coinbase, nil
+}
+
+func (b *ethAPIBackend) GetEVM(ctx context.Context, msg *core.Message, sdb *state.StateDB, hdr *types.Header, cfg *vm.Config, bCtx *vm.BlockContext) *vm.EVM {
```
StephenButtolph: In `eth_accessList`, the API provides `bCtx == nil`. Can we on of:
- handle it gracefully like they do in libevm (will need to provide a `GetHeader` function somehow)
```golang
var context vm.BlockContext
if bCtx != nil {
	context = *bCtx
} else {
	context = core.NewEVMBlockContext(header, b.eth.BlockChain(), nil)
}
```
- Put a TODO so we know we need to handle it when we go to test that RPC?

sae/rpc_stateful.go
```diff
@@ -0,0 +1,83 @@
...
+	txCtx := vm.TxContext{
+		Origin:   msg.From,
+		GasPrice: big.NewInt(1),
+	}
+	return vm.NewEVM(*bCtx, txCtx, sdb, b.ChainConfig(), *cfg)
+}
+
+func (b *ethAPIBackend) StateAndHeaderByNumber(ctx context.Context, num rpc.BlockNumber) (*state.StateDB, *types.Header, error) {
+	return b.StateAndHeaderByNumberOrHash(ctx, rpc.BlockNumberOrHashWithNumber(num))
+}
```
StephenButtolph: Fun geth fact, this function is never called.

ARR4N: Nothing is fun about RPCs

sae/rpc_test.go
```diff
@@ -952,3 +953,101 @@ func TestDebugNamespace(t *testing.T) {
...
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			be := sut.rawVM.apiBackend
+			gotNum, gotHash, err := be.resolveBlockNumberOrHash(tt.nOrH)
+			if !errors.Is(err, tt.wantErr) || gotNum != tt.wantNum || gotHash != tt.wantHash {
+				t.Errorf("%T.resolveBlockNumberOrHash(%+v) got (%d, %#x, %v); want (%d, %#x, Is(%v))", be, tt.nOrH, gotNum, gotHash, err, tt.wantNum, tt.wantHash, tt.wantErr)
+			}
```
StephenButtolph: For someone that basically just looks at the line number and not the actual error text (me), stuffing all of these checks onto one line feels counter-productive. Is there a reason we don't split this up?
```suggestion
			require.ErrorIs(t, err, tt.wantErr)
			assert.Equal(t, tt.wantNum, gotNum)
			assert.Equal(t, tt.wantHash, gotHash)
```

My guess is that you wanted to avoid duplicating the error message over and over again... So 🤷

StephenButtolph (approved):

PR #169, opened 2026-02-17
feat: immediate receipt issuance

ARR4N:
The `geth` APIs fetch individual transaction results by loading the entire block's receipts, which forces all senders to wait until the whole block has finished execution. This PR introduces a cache of recent receipts, and intercepts the `GetTransactionReceipt()` RPC method, serving from the cache if it holds the data. This allows for receipt issuance immediately after _transaction_ execution, not after _block_ execution. The cache is cleared at the same time as the respective block is ejected from `VM` memory.

To further reduce polling of `GetTransactionReceipt()`, the caching mechanism is aware of pending (accepted but not executed) transactions and will block until they are executed or context cancellation occurs. The test case in which a pending transaction returns `nil, nil` for receipt requests is therefore removed.

The `saexec.syncMap` type is a thread-safe key-value store that aims to reduce lock contention by spreading reads and writes over independent buckets. Without this, every enqueued block (execution thread) and every cache clearance (GC) would block receipt storage / retrieval. Both scenarios only hold 1/256 locks at any time to avoid starving the RPCs.

Although existing receipt tests hit the current path, an explicit unit test demonstrates that blocking of a transaction doesn't block earlier receipt availability.

Closes ava-labs/avalanchego#5252

### DIFF
--- a/saexec/receipts.go
+++ b/saexec/receipts.go
@@ -109,25 +109,25 @@ func newSyncMap[K ~[32]byte, V any]() *syncMap[K, V] {
 // number of buckets, not the number of keys. To achieve this, the caller MUST
 // consider the order in which keys are passed to `fn` to be undefined.
 func (m *syncMap[K, V]) writeMany(fn func(map[K]V, K), ks ...K) {
-	if len(ks) == 0 {
-		return
-	}
 	slices.SortFunc(ks, func(a, b K) int {
 		return cmp.Compare(a[0], b[0])
 	})
 
-	var locked *bucket[K, V]
-	for _, k := range ks {
-		if b := &m.buckets[k[0]]; b != locked {
-			if locked != nil {
-				locked.Unlock()
-			}
-			locked = b
-			locked.Lock()
+	var b *bucket[K, V]
+	unlock := func() {
+		if b != nil {
+			b.Unlock()
 		}
-		fn(locked.data, k)
 	}
-	locked.Unlock()
+	for _, k := range ks {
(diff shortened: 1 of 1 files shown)


cache/cache.go
StephenButtolph: Is there a "happens before" relationship for this data structure?

While the _usage_ of this cache makes sense... The implementation seems unusual.

For example:
```golang
func TestCacheConsistency(t *testing.T) {
	c := NewUniformlyKeyed[common.Hash, int]()

	var wg sync.WaitGroup
	wg.Add(2)

	go func() {
		defer wg.Done()
		c.Delete(common.Hash{0}, common.Hash{1})
	}()
	go func() {
		defer wg.Done()
		c.Store(common.Hash{0}, 0)
		c.Store(common.Hash{1}, 1)
	}()

	/*
		Possible executions:

		1:
			Delete(0, 1)
			Store(0)
			Store(1)

			State: (0, 1)

		2:
			Store(0)
			Delete(0, 1)
			Store(1)

			State: (1)

		3:
			Store(0)
			Store(1)
			Delete(0, 1)

			State: ()

		So we should never expect a cache state to exist where only (0) is in
		the cache
	*/

	wg.Wait()
	if _, ok := c.Load(common.Hash{0}); !ok {
		return
	}
	if _, ok := c.Load(common.Hash{1}); !ok {
		t.Errorf("cache contains 0 but not 1")
	}
}
```

I would have expected this to be a valid test, but it can currently fail because `Delete` is not actually an atomic operation.

StephenButtolph (approved):

PR #175, opened 2026-02-17
test: fix wait until executed flakes definitively

### LINKED ISSUE #172: Fix `TestEthGetters` flake
See https://github.com/ava-labs/strevm/actions/runs/22114333831/job/63918376186?pr=168

JonathanOppenheimer:
Closes #172.

Here, we sync the mempool in `runConsensusLoop` so every test that builds a block automatically has a fully-processed mempool. This fixes flakes of the the type surfaced with `TestTxTypeSupport` where `BuildBlock` raced with async mempool ingestion and only included 1 of 3 submitted transactions (see https://github.com/ava-labs/strevm/actions/runs/21882243041/job/63167673465?pr=136).

This is necessary because `SendTx` calls `pool.Add` with `sync=false`, so transactions are queued for async processing. Without syncing, `BuildBlock` won't wait for the pool to ingest all the submitted transactions, and we had to manually call `syncMempool` in the tests, and if you forget you'd  get an intermittent flake. So we can eliminate this entirely by moving the sync into `runConsensusLoop`. We can also remove the now unnecessary `syncMempool` calls 

Lastly, we also wait for the last accepted block to finish executing in newSUT cleanup, preventing potential flakes from the executor still running during shutdown.

Hopefully no more flakes!

### DIFF
--- a/sae/recovery_test.go
+++ b/sae/recovery_test.go
@@ -59,7 +59,7 @@ func TestRecoverFromDatabase(t *testing.T) {
 				Gas:      params.TxGas + params.CreateGas + params.TxDataNonZeroGasFrontier + rng.Uint64N(2e6),
 				GasPrice: big.NewInt(100),
 			}))
-			}
+		}
 
 		vmTime.advance(850 * time.Millisecond)
 		b := src.runConsensusLoop(t, src.lastAcceptedBlock(t))


sae/vm_test.go
```diff
@@ -137,6 +137,7 @@ func newSUT(tb testing.TB, numAccounts uint, opts ...sutOption) (context.Context
 	snow := adaptor.Convert(vm)
 	tb.Cleanup(func() {
 		ctx := context.WithoutCancel(tb.Context())
+		require.NoError(tb, vm.last.accepted.Load().WaitUntilExecuted(ctx), "WaitUntilExecuted()")
```
```suggestion
		require.NoError(tb, vm.last.accepted.Load().WaitUntilExecuted(ctx), "{last-accepted block}.WaitUntilExecuted()")
```

PR #183, opened 2026-02-18
feat: Remaining ChainVM methods

alarso16:
Add GetBlock and GetBlockIDByHeight with basic tests. The refactoring necessary to expose the settling action is really gross, so we might just need to simplify some of that eventually (very panic-prone). Additionally, whether a block is synchronous or not is not persisted, so it cannot be loaded.

### DIFF
--- a/blocks/settlement.go
+++ b/blocks/settlement.go
@@ -84,7 +84,7 @@ func (b *Block) markSettled(lastSettled *atomic.Pointer[Block]) error {
 // [Block.SetAsHeadBlock], which MUST be done by the caller, i.f.f. the chain
 // has not yet commenced asynchronous execution.
 //
-// TODO(arr4n) refactor to avoid DB writes on every startup.
+// TODO(arr4n) refactor to avoid requiring DB writes.
 func (b *Block) MarkSynchronous(hooks hook.Points, db ethdb.Database, xdb saedb.ExecutionResults, excessAfter gas.Gas) error {
 	ethB := b.EthBlock()
 	// Receipts of a synchronous block have already been "settled" by the block


blocks/cmpopt.go
```diff
@@ -37,18 +36,17 @@ func CmpOpt() cmp.Option {
...
-func (e *executionResults) equalForTests(f *executionResults) bool {
-	fn := cmputils.WithNilCheck(func(e, f *executionResults) bool {
-		return true &&
-			e.byGas.Rate() == f.byGas.Rate() &&
+func executionResultsCmpOpt() cmp.Option {
+	fn := func(e, f *executionResults) bool {
+		res := e.byGas.Rate() == f.byGas.Rate() &&
 			e.byGas.Compare(f.byGas.Time) == 0 &&
 			e.receiptRoot == f.receiptRoot &&
-			saetest.MerkleRootsEqual(e.receipts, f.receipts) &&
```
alarso16: This shouldn't be checked, as it's not even persisted in execution results.

sae/vm_test.go
```diff
@@ -343,21 +345,32 @@ func (s *SUT) requireInMempool(tb testing.TB, txs ...common.Hash) {
...
-	return s.runConsensusLoop(tb, s.lastAcceptedBlock(tb))
+	return s.runConsensusLoop(tb, s.lastAcceptedBlock(tb), true)
 }
 
 // runConsensusLoop syncs the mempool, sets the preference to the specified
 // block, then builds, verifies, accepts, and returns the new block. It does NOT
 // wait for it to be executed; to do this automatically, set the [VM] to
 // [snow.Bootstrapping].
-func (s *SUT) runConsensusLoop(tb testing.TB, preference *blocks.Block) *blocks.Block {
+func (s *SUT) runConsensusLoop(tb testing.TB, preference *blocks.Block, accept bool) *blocks.Block {
```
alarso16: Not my best work... would it be better to put this optional or something? Or just write separate functions

saexec/saexec.go
```diff
@@ -137,3 +137,8 @@ func (e *Executor) StateCache() state.Database {
 func (e *Executor) LastExecuted() *blocks.Block {
 	return e.lastExecuted.Load()
 }
+
+// ExecutionResults fills a block with any associated execution results.
+func (e *Executor) ExecutionResults(b *blocks.Block) error {
+	return b.RestoreExecutionArtefacts(e.db, e.xdb, e.chainConfig)
+}
```
StephenButtolph: Do we need to add this? I think we do this already outside of the executor.

sae/blocks.go
```diff
@@ -358,24 +359,60 @@ func (vm *VM) VerifyBlock(ctx context.Context, bCtx *block.Context, b *blocks.Bl
...
+	// Will find all verified/accepted blocks not settled.
+	b, ok := vm.blocks.Load(hash)
+	if ok {
+		return b, nil
+	}
+
+	// If not in memory, then it must be settled and on disk.
+	ethB := readByHash(vm, hash, (*blocks.Block).EthBlock, rawdb.ReadBlock)
+	if ethB == nil {
+		return nil, fmt.Errorf("%w: block %s not found on disk", database.ErrNotFound, id)
```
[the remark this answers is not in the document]
alarso16: Gross. But I'll fix it

sae/vm_test.go
alarso16: It is, this seems better

sae/blocks.go
```diff
@@ -364,18 +365,67 @@ func canonicalBlock(db ethdb.Database, num uint64) (*types.Block, error) {
...
+func (vm *VM) GetBlock(_ context.Context, id ids.ID) (*blocks.Block, error) {
+	var _ snowman.Block // protect the input to allow comment linking
+
+	// Check for any blocks that have been verified, but not yet released after
+	// settling.
+	// readByHash does also load from `vm.blocks`, but it will not contain the full
+	// exeuction data necessary for returning, so it must be called separately.
+	hash := common.Hash(id)
+	b, ok := vm.blocks.Load(hash)
+	if ok {
```
ARR4N:
```suggestion
	if b, ok := vm.blocks.Load(hash); ok {
```

sae/blocks.go
```diff
@@ -364,18 +365,67 @@ func canonicalBlock(db ethdb.Database, num uint64) (*types.Block, error) {
...
+		return nil, err
+	}
 	return b, nil
 }
 
 // GetBlockIDAtHeight returns the accepted block at the given height, or
 // [database.ErrNotFound].
-func (vm *VM) GetBlockIDAtHeight(context.Context, uint64) (ids.ID, error) {
-	return ids.Empty, errUnimplemented
+func (vm *VM) GetBlockIDAtHeight(_ context.Context, height uint64) (ids.ID, error) {
```
ARR4N:
```suggestion
func (vm *VM) GetBlockIDAtHeight(ctx context.Context, height uint64) (ids.ID, error) {
```

blocks/settlement.go
```diff
@@ -105,7 +101,7 @@ func (b *Block) MarkSynchronous(hooks hook.Points, db ethdb.Database, xdb saedb.
 		return err
 	}
 	b.synchronous = true
-	return b.markSettled(nil)
+	return b.MarkSettled(nil)
```
ARR4N: This used the internal one to allow for a `nil` pointer to be passed. Please revert the change.

I suspect that you made these changes to get around having to pass a single-use `new(atomic.Pointer[blocks.Block])` when calling the function. The correct approach is to either pass the blackhole pointer (I agree, it's gross) or to introduce a `NewSettledBlock()` function.

blocks/blockstest/chain.go
```diff
@@ -31,10 +31,10 @@ type ChainBuilder struct {
 
 // NewChainBuilder returns a new ChainBuilder starting from the provided block,
 // which MUST NOT be nil.
-func NewChainBuilder(config *params.ChainConfig, genesis *blocks.Block, defaultOpts ...ChainOption) *ChainBuilder {
+func NewChainBuilder(config *params.ChainConfig, parent *blocks.Block, defaultOpts ...ChainOption) *ChainBuilder {
 	c := &ChainBuilder{
 		config: config,
-		chain:  []*blocks.Block{genesis},
+		chain:  []*blocks.Block{parent},
```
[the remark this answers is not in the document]
alarso16: I meant to restore this - I tried using it to make a well-formed block at one point for the race-y test, but it was for some reason not well-formed... I got RLP errors. Anyway, it can be reverted (although it doesn't have to be the genesis block)

blocks/settlement.go
```diff
@@ -83,6 +83,8 @@ func (b *Block) markSettled(lastSettled *atomic.Pointer[Block]) error {
 // Unlike [Block.MarkExecuted], MarkSynchronous does not call
 // [Block.SetAsHeadBlock], which MUST be done by the caller, i.f.f. the chain
 // has not yet commenced asynchronous execution.
+//
+// TODO(arr4n) refactor to avoid DB writes on every startup.
```
StephenButtolph: Not just done on startup anymore
```suggestion
// TODO(arr4n) refactor to avoid requiring DB writes.
```

PR #213, opened 2026-02-20
Suggestions for PR 161

StephenButtolph:
I started reviewing, #161, but then started making a suggestions PR and then I got pretty carried away.

I'll be documenting my rationale in-line in this PR

### DIFF
--- a/gasprice/estimator_test.go
+++ b/gasprice/estimator_test.go
@@ -349,6 +349,15 @@ func TestFeeHistory(t *testing.T) {
 		NextGasTime: gastime.New(time.Now(), 1, math.MaxUint64),
 	}
 
+	newBlock := func(sut *SUT, txs ...*types.Transaction) *blocks.Block {
+		t.Helper()
+		return sut.newBlock(t, 0, bounds, txs...)
+	}
+	newTx := func(sut *SUT, gas, price uint64) *types.Transaction {
+		t.Helper()
+		return sut.newTx(t, gas, price)
+	}
+
 	type args struct {
 		numBlocks   uint64
 		lastBlock   rpc.BlockNumber
@@ -363,7 +372,7 @@ func TestFeeHistory(t *testing.T) {
 	}
 	tests := []struct {
 		name   string
-		blocks func(*testing.T, *SUT) []*blocks.Block
+		blocks func(*SUT) []*blocks.Block
 		args   args
 		want   results
 	}{
@@ -429,9 +438,9 @@ func TestFeeHistory(t *testing.T) {
 		},
 		{
 			name: "query_genesis",
-			blocks: func(t *testing.T, sut *SUT) []*blocks.Block {
+			blocks: func(sut *SUT) []*blocks.Block {
 				return []*blocks.Block{
-					sut.newBlock(t, 0, bounds, sut.newTx(t, 21_000, nAVAX)),
+					newBlock(sut, newTx(sut, 21_000, nAVAX)),
 				}
 			},
 			args: args{
@@ -451,9 +460,9 @@ func TestFeeHistory(t *testing.T) {
 		},
 		{
 			name: "query_latest",
-			blocks: func(t *testing.T, sut *SUT) []*blocks.Block {
+			blocks: func(sut *SUT) []*blocks.Block {
 				return []*blocks.Block{
-					sut.newBlock(t, 0, bounds, sut.newTx(t, 21_000, nAVAX)),
+					newBlock(sut, newTx(sut, 21_000, nAVAX)),
 				}
 			},
 			args: args{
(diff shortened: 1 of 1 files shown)


gasprice/block.go
StephenButtolph: This file was renamed from `fee_info_provider.go`

gasprice/block.go
```diff
@@ -0,0 +1,137 @@
...
+
+// tipPercentiles computes the gas-weighted tip at each requested percentile.
+//
+// Because block builders sequence transactions without executing them in SAE,
+// we accumulate gas limits, not the gas charged.
+func (b *block) tipPercentiles(percentiles []float64) []*big.Int {
+	out := make([]*big.Int, len(percentiles))
+	if len(b.txs) == 0 {
+		for i := range out {
+			out[i] = common.Big0
```
StephenButtolph: In the prior code we returned `new(big.Int)`, but returned `tx.tip` without a copy... So I try to avoid the memory allocation here.

gasprice/fee_info_provider.go
```diff
@@ -1,183 +0,0 @@
...
-	}
-	if block == nil {
-		return nil, nil
-	}
-	return f.addBlock(block), nil
-}
-
-// populateCache fills the cache with up to size blocks ending at last accepted.
-// Assumes size is greater than zero.
-func (f *feeInfoProvider) populateCache(size uint64) error {
```
StephenButtolph: I removed the explicit cache population on startup. Our nodes will normally bootstrap some recent blocks during startup... So I don't think this is actually needed normally... Even if it _is_ hit, the API will still work (albeit slowly for the first call).

gasprice/config.go
```diff
@@ -1,97 +0,0 @@
...
-		Percentile:          60,
-		MaxLookbackSeconds:  uint64(80),
-		MaxCallBlockHistory: 2048,
-		// Default MaxBlockHistory is chosen to be a value larger than the required
-		// fee lookback window that MetaMask uses (20k blocks).
-		MaxBlockHistory: 25_000,
-		MaxPrice:        big.NewInt(150 * params.GWei),
-		MinPrice:        big.NewInt(1 * params.Wei),
-	}
-}
```
StephenButtolph: Moved the defaults to `setDefaults`. I prefer the `setDefaults` approach so that the user of the package can just set whichever fields they care about in the config. Rather than needing to get a populated config and then override the values.

gasprice/estimator.go
```diff
@@ -10,278 +10,299 @@ import (
...
-// Backend abstracts the chain data needed by the Estimator.
+// Backend that the [Estimator] depends on for chain data.
 type Backend interface {
 	ResolveBlockNumber(bn rpc.BlockNumber) (uint64, error)
 	BlockByNumber(ctx context.Context, number rpc.BlockNumber) (*types.Block, error)
-	SubscribeChainAcceptedEvent(ch chan<- *types.Block) event.Subscription
-	// NextBaseFeeUpperBound returns the worst-case (upper-bound) base fee for
-	// the block after the last accepted head. Returns nil when unavailable.
-	NextBaseFeeUpperBound(ctx context.Context) *big.Int
+	SubscribeChainHeadEvent(ch chan<- core.ChainHeadEvent) event.Subscription
```
StephenButtolph: I opted for only preemptively caching executed blocks rather than upon accept and leaving a TODO rather than modifying the executor again.

gasprice/estimator.go
```diff
@@ -10,278 +10,299 @@ import (
...
-	// Slow path: acquire exclusive lock to recompute the price.
-	// Re-check after acquiring the lock because another goroutine may have
-	// already computed the price for this head while we were waiting.
 	e.lastLock.Lock()
 	defer e.lastLock.Unlock()
+
 	lastNumber, lastPrice = e.lastNumber, e.lastPrice
-	if headNumber == lastNumber {
-		return new(big.Int).Set(lastPrice), nil
+	if headNumber <= lastNumber {
```
StephenButtolph: Minor fix here, if we had concurrent calls, it was actually possible for us to _decrease_ the `lastNumber`.. Which was definitely unexpected... Now we only update the gpo for higher heights.

gasprice/estimator.go
```diff
@@ -10,278 +10,307 @@ import (
...
-	// Truncate to the actually processed range.
-	baseFee = baseFee[:filled]
-	gasUsedRatio = gasUsedRatio[:filled]
-	if len(rewardPercentiles) != 0 {
-		reward = reward[:filled]
+	if last == head {
+		bounds := headBlock.WorstCaseBounds()
+		if bounds == nil {
+			return common.Big0, nil, nil, nil, errMissingWorstCaseBounds
+		}
```
StephenButtolph: Rather than falling back to a different (incorrect) base fee, we should just error.

gasprice/estimator_test.go
StephenButtolph: This file seemed to really follow the coreth/geth tests. I updated it to follow a more sae-ie pattern.

gasprice/block.go
ceyonur: Is `common.Big0` safe to return? I don't see where we copy this and modifying this could end-up a disaster (yes happened before lol)

gasprice/block.go
```diff
@@ -0,0 +1,137 @@
...
+func (b *blockCache) getBlock(ctx context.Context, n uint64) *block {
+	if blk, ok := b.cache.Get(n); ok {
+		return blk
+	}
+
+	blk, err := b.backend.BlockByNumber(ctx, rpc.BlockNumber(n)) //nolint:gosec // block numbers were previously resolved
+	if err != nil {
+		b.log.Error("fetching BlockByNumber",
+			zap.Uint64("number", n),
+			zap.Error(err),
```
ceyonur: what happen if ctx is cancelled here rather than err in db read. I think in that case we would log an error unnecessarily + it's indistinguishable from the caller side if it's missing block or ctx cancelled, right?

gasprice/estimator_test.go
```diff
@@ -7,525 +7,445 @@ import (
...
-
-var backendBaseFee = big.NewInt(875_000_000)
-
-// testBlockGen is a simplified block generator for tests, replacing core.BlockGen.
-type testBlockGen struct {
-	baseFee *big.Int
-	txs     types.Transactions
-	nonces  map[common.Address]uint64
+func TestMain(m *testing.M) {
+	goleak.VerifyTestMain(m, goleak.IgnoreCurrent())
```
ceyonur: ooo this looks useful. I wonder if we can lintify this?

gasprice/estimator_test.go
```diff
@@ -7,525 +7,445 @@ import (
...
-				Nonce:     b.TxNonce(addr),
-				To:        &common.Address{},
-				Gas:       params.TxGas,
-				GasFeeCap: feeCap,
-				GasTipCap: txTip,
-				Data:      []byte{},
-			})
-			b.AddTx(tx)
-		}
+	newBlock := func(time uint64, txs ...*types.Transaction) *blocks.Block {
```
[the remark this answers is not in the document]
ceyonur: I tried to simplify and make these tests stateless

PR #216, opened 2026-02-20
perf: improve worstcase state initialization

### LINKED ISSUE #185: Snapshot support in worst-case state initialization
See: https://github.com/ava-labs/strevm/blob/5faf8141974efdc39dca07b179f74c18db7a8718/worstcase/state.go#L76

`state.New()` is called with `nil` for the snapshot parameter when creating worst-case state from a settled block's post-execution state root.

Worth benchmarking whether passing a snapshot here would meaningfully speed up state reads during worst-case validation. If not, `nil` would be fine.

JonathanOppenheimer:
Closes #185 

@StephenButtolph asked 

> // TODO: Should we be providing snapshots here?

The answer seems to be yes! I added a benchmark, `BenchmarkApplyTxWithSnapshot`, which shows consistent performance improvements with snapshots supported: 

`go test ./worstcase/... -run=^$ -bench=BenchmarkApplyTxWithSnapshot -benchtime=100ms -timeout 600s 2>&1` 

| Accounts | Without Snapshot | With Snapshot | Speedup |                                                                                                                      
  |----------|-----------------|---------------|---------|
  | 100      | 5,087 ns/op     | 3,657 ns/op   | 1.39x   |                                                                                                                       
  | 1,000    | 5,711 ns/op     | 3,661 ns/op   | 1.56x   |                                                                                                                     
  | 10,000   | 6,350 ns/op     | 3,861 ns/op   | 1.64x   |

This PR is pretty simply and updates `worstcase.NewState` to accept a snapshot tree and pass it to `state.New(...)` instead of always using nil.

### DIFF
--- a/worstcase/state_benchmark_test.go
+++ b/worstcase/state_benchmark_test.go
@@ -33,7 +33,7 @@ func BenchmarkApplyTxWithSnapshot(b *testing.B) {
 
 			snaps, err := snapshot.New(
 				snapshot.Config{
-					AsyncBuild: true,
+					AsyncBuild: false,
 					CacheSize:  saexec.SnapshotCacheSizeMB,
 				},
 				sut.db, sut.stateCache.TrieDB(), sut.genesis.PostExecutionStateRoot(),
--- a/saexec/saexec.go
+++ b/saexec/saexec.go
@@ -52,11 +52,9 @@ type Executor struct {
 	// [Executor.execute] and [Executor.Close]. Callers MUST treat
 	// values returned from [Executor.SnapshotTree] as read-only.
 	//
-	// [snapshot.Tree] is safe for concurrent read access.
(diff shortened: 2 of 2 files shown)


worstcase/state_benchmark_test.go
```diff
@@ -0,0 +1,69 @@
...
+			sut := newSUT(b, saetest.MaxAllocFor(wallet.Addresses()...))
+
+			tx := wallet.SetNonceAndSign(b, 0, &types.LegacyTx{
+				To:       &common.Address{},
+				Gas:      params.TxGas,
+				GasPrice: big.NewInt(1),
+			})
+
+			snapTree, err := snapshot.New(
+				snapshot.Config{CacheSize: saexec.SnapshotCacheSizeMB},
```
ARR4N: Add `AsyncBuild: false` so we know that it's in a ready state by the time its constructor returns.

PR #217, opened 2026-02-21
feat: `Executor` cache of recent block headers + `VM` fallback

ARR4N:
The `BLOCKHASH` op code requires the `core.ChainContext` to provide access to the last 256 block headers. This PR adds an LRU into `saexec`'s `chainContext` implementation and also provides full historical retrieval for all headers via the `VM`. To achieve the latter, the existing `blocks.Source` type has two similar types added that don't require a full `blocks.Block` to be constructed.

### DIFF
--- a/blocks/blockstest/chain.go
+++ b/blocks/blockstest/chain.go
@@ -80,7 +80,7 @@ func (cb *ChainBuilder) NewBlock(tb testing.TB, txs []*types.Transaction, opts .
 
 	last := cb.Last()
 	eth := NewEthBlock(last.EthBlock(), txs, allOpts.eth...)
-	b := NewBlock(tb, eth, last, nil, allOpts.sae...) // TODO(arr4n) support last-settled blocks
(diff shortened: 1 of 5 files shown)


saexec/execution.go
```diff
@@ -81,6 +81,13 @@ func (e *Executor) processQueue() {
 func (e *Executor) execute(b *blocks.Block, logger logging.Logger) error {
 	logger.Debug("Executing block")
 
+	defer func() {
+		// The current execution already has access via the [vm.BlockContext]
+		// whereas the [core.ChainContext] provides access to historical blocks
+		// for the BLOCKHASH op code.
+		e.chainContext.recent.Put(b.NumberU64(), b.Header())
+	}()
```
StephenButtolph: Why is this done in a separate func?
```suggestion
	// The current execution already has access via the [vm.BlockContext]
	// whereas the [core.ChainContext] provides access to historical blocks for
	// the BLOCKHASH op code.
	defer e.chainContext.recent.Put(b.NumberU64(), b.Header())
```

PR #220, opened 2026-02-23
refactor!: remove `proxytime.Time.SetRateInvariants()` and expose `Scale()` instead

ARR4N:
As general-purpose invariants, it would have been inappropriate for `proxytime.Time` to implement specific overflow behaviour (e.g. capping) so it returned an error. This was the incorrect abstraction so the invariants have been removed in favour of exposing the `Scale()` method for external use. `gastime.Time` is now responsible for scaling excess, and caps it at `2^64 - 1` instead of returning an error, which makes scaling error-free.

The remaining excess overflow (when `Tick()`ing) is deferred for a follow-up PR.

### DIFF
--- a/blocks/time_test.go
+++ b/blocks/time_test.go
@@ -11,6 +11,7 @@ import (
 	"github.com/ava-labs/avalanchego/vms/components/gas"
 	"github.com/ava-labs/libevm/core/types"
 	"github.com/google/go-cmp/cmp"
+	"github.com/stretchr/testify/require"
 
 	"github.com/ava-labs/strevm/gastime"
 	"github.com/ava-labs/strevm/hook/hookstest"


gastime/gastime.go
```diff
@@ -164,17 +156,28 @@ func (tm *Time) BaseFee() *uint256.Int {
...
 }
 
-// SetRate changes the gas rate per second, rounding down the argument if it is
-// not a multiple of [TargetToRate]. See [Time.SetTarget] re potential error(s).
-func (tm *Time) SetRate(r gas.Gas) error {
-	return tm.TimeMarshaler.SetRate(roundRate(r))
+// SetRate is equivalent to [Time.SetTarget] after (integer) division of `r` by
+// [TargetToRate].
+func (tm *Time) SetRate(r gas.Gas) {
+	tm.SetTarget(r / TargetToRate)
```
ceyonur: why are we not rounding this anymore?

proxytime/proxytime.go
StephenButtolph: Now that we don't have the `rateInvariants`, should we remove the `nocopy` and just copy the struct directly? Aka:
```golang
// Time represents an instant in time, its passage measured by an arbitrary unit
// of duration. It is not thread safe nor is the zero value valid.
type Time[D Duration] struct {
	seconds uint64 `canoto:"uint,1"`
	// invariant: fraction < hertz
	fraction D `canoto:"uint,2"`
	hertz    D `canoto:"uint,3"`

	canotoData canotoData_Time
}

// Clone returns a copy of the time.
func (tm *Time[D]) Clone() *Time[D] {
	c := *tm
	return &c
}
```

With this simplification, it might even make sense to remove `Clone`. But I think it's cleanest to leave it.

proxytime/proxytime.go
```diff
@@ -171,52 +167,22 @@ func ConvertMilliseconds[D Duration](rate D, ms uint64) (sec uint64, _ Fractiona
...
-	}
-	for i, v := range tm.rateInvariants {
-		*v = scaled[i]
-	}
-
+func (tm *Time[D]) SetRate(hertz D) {
+	// If this happens then there is a bug in the implementation. The
+	// invariant that `tm.fraction < tm.hertz` makes overflow impossible as
+	// the scaled fraction will be less than the new rate.
+	frac, _ := tm.Scale(tm.fraction, hertz)
```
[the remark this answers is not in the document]
StephenButtolph: tbh, I'd prefer the brittle assumption rather than re-adding the `TimeMarshaler` type.

PR #221, opened 2026-02-23
fix: keep txs in `TestEthGetters`

ARR4N:
Reverts accidental removal of txs in `TestEthGetters` in #218. It's not ideal having actual test assertions being separated out but a full refactor is beyond the scope of this PR.


StephenButtolph (approved):
I don't understand the rationale for moving these tests around (and it made the diff pretty difficult to actually review).

That said, I locally replicated the PR and the change does seem to fix the removed tx testing.. So lgtm

PR #233, opened 2026-02-24
feat(hook): add allowlist hooks for pre-block transaction and contract filtering

powerslider:
Extend `hook.Points` with `CanExecuteTransaction`, mirroring `params.RulesAllowlistHooks` so consumers can use a single concrete type for both SAE and `libevm` hooks. Call `CanExecuteTransaction` via hooks in `worstcase.ApplyTx()` to exclude hook-rejected transactions before block inclusion, preventing fatal errors in the execution
loop.

- Add `CanExecuteTransaction` to `hook.Points` and `hookstest.Stub`.
- Call `CanExecuteTransaction` in `worstcase.State.ApplyTx()` instead of through `libevm` rules hooks.
- Remove dead rules field from `worstcase.State`.
- Add `TestCanExecuteTransactionHook` in `worstcase` package verifying blocked senders are excluded and state is not modified.
- Add `TestCanCreateContractSoftError` confirming `CanCreateContract` rejection produces a failed receipt, not a fatal crash.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -27,7 +27,7 @@ require (
 	github.com/btcsuite/btcd/btcec/v2 v2.3.5 // indirect
 	github.com/cenkalti/backoff/v5 v5.0.2 // indirect
 	github.com/chainguard-dev/git-urls v1.0.2 // indirect
-	github.com/cloudflare/circl v1.6.1 // indirect
+	github.com/cloudflare/circl v1.6.3 // indirect
 	github.com/cockroachdb/errors v1.9.1 // indirect
 	github.com/cockroachdb/logtags v0.0.0-20230118201751-21c54148d20b // indirect
 	github.com/cockroachdb/pebble v0.0.0-20230928194634-aa077af62593 // indirect
--- a/go.sum
+++ b/go.sum
@@ -69,8 +69,8 @@ github.com/chzyer/logex v1.1.10/go.mod h1:+Ywpsq7O8HXn0nuIou7OrIPyXbp3wmkHB+jjWR
 github.com/chzyer/readline v0.0.0-20180603132655-2972be24d48e/go.mod h1:nSuG5e5PlCu98SY8svDHJxuZscDgtXS6KTTbou5AhLI=
 github.com/chzyer/test v0.0.0-20180213035817-a1ea475d72b1/go.mod h1:Q3SI9o4m/ZMnBNeIyt5eFwwo7qiLfzFZmjNmxjkiQlU=
 github.com/client9/misspell v0.3.4/go.mod h1:qj6jICC3Q7zFZvVWo7KLAzC3yx5G7kyvSDkc90ppPyw=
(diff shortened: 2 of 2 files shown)


worstcase/state.go
```diff
@@ -56,7 +56,6 @@ type State struct {
 
 	baseFee             *uint256.Int
 	curr                *types.Header
-	rules               params.Rules
```
ARR4N: Good catch, thank you.

hook/hook.go
```diff
@@ -56,6 +57,12 @@ type Points interface {
 	// These operations will be performed during both worst-case and actual
 	// execution.
 	EndOfBlockOps(*types.Block) []Op
+	// CanCreateContract mirrors [params.RulesAllowlistHooks.CanCreateContract]
+	// so that consumers can use a single concrete type for both SAE and libevm hooks.
+	CanCreateContract(*libevm.AddressContext, uint64, libevm.StateReader) (uint64, error)
```
```suggestion
```

Because you never use it and that your test proves it's safe to do so.

worstcase/state_test.go
```diff
@@ -615,3 +618,75 @@ func TestStartBlockQueueFullDueToTargetChanges(t *testing.T) {
...
+
+	config := sut.config
+	signer := types.NewCancunSigner(config.ChainID)
+
+	tests := []struct {
+		name      string
+		key       *ecdsa.PrivateKey
+		addr      common.Address
+		wantErr   error
+		wantNonce uint64
```
ARR4N: `wantNonceAfter`

sae/vm_test.go
```diff
@@ -577,6 +580,43 @@ func TestIntegration(t *testing.T) {
...
+	ctx, sut := newSUT(t, 1)
+
+	stub := &libevmhookstest.Stub{
+		CanCreateContractFn: func(_ *libevm.AddressContext, _ uint64, _ libevm.StateReader) (uint64, error) {
+			return 0, errContractCreationBlocked
+		},
+		MinimumGasConsumptionFn: hook.MinimumGasConsumption,
+	}
+	libevmExtras := stub.Register(t)
+	libevmExtras.ChainConfig.Set(sut.rawVM.exec.ChainConfig(), stub)
```
```suggestion
	stub.Register(t)
```

Setting the extra in the config is only necessary if it's a data payload. In this case it's a hook. Think of it as setting a field in a struct vs defining a method.

sae/vm_test.go
```diff
@@ -577,6 +580,43 @@ func TestIntegration(t *testing.T) {
...
+
+	const gasLimit uint64 = 100_000
+	tx := sut.wallet.SetNonceAndSign(t, 0, &types.DynamicFeeTx{
+		To:        nil, // contract creation
+		Gas:       gasLimit,
+		GasFeeCap: big.NewInt(1),
+	})
+	sut.mustSendTx(t, tx)
+
+	b := sut.runConsensusLoop(t)
```
```suggestion

	b := sut.runConsensusLoop(t, tx)
```

ARR4N (approved):

PR #234, opened 2026-02-24
fix: populate gas used properly

### LINKED ISSUE #193: Should `hook.BlockBuilder` populate `GasUsed`?
See: https://github.com/ava-labs/strevm/blob/5faf8141974efdc39dca07b179f74c18db7a8718/sae/blocks.go#L243-L244

`GasUsed` is set from `state.GasUsed()` during block building, but hook operations might consume additional gas not captured here.

Check whether `hook.Op.Gas` actually contributes meaningful gas that's being missed. If so, the block builder should account for it. I also noticed this yesterday and talked to @StephenButtolph about it.

JonathanOppenheimer:
Closes #193. 

The `[hook.BlockBuilder]`  should populate `[types.Header.GasUsed`] so that [hook.Op.Gas] can be included.  This will make the reported GasUsed more representative of reality because atomics use part of the gas limit.

### DIFF
--- a/hook/hook.go
+++ b/hook/hook.go
@@ -76,8 +76,9 @@ type BlockBuilder interface {
 	//
 	// The returned header MUST have [types.Header.ParentHash],
 	// [types.Header.Number] and [types.Header.Time] set appropriately.
-	// [types.Header.Root], [types.Header.GasLimit], and [types.Header.BaseFee]
-	// will be ignored and overwritten. Any other fields MAY be set as desired.
+	// [types.Header.Root], [types.Header.GasLimit], [types.Header.BaseFee], and
+	// [types.Header.GasUsed] will be ignored and overwritten. Any other fields
+	// MAY be set as desired.
 	//
 	// SAE always uses this method instead of directly constructing a header, to
 	// ensure any libevm header extras are properly populated.


sae/blocks.go
```diff
@@ -241,8 +241,33 @@ func (vm *VM) buildBlock(
...
+	for _, b := range settling {
+		receipts = append(receipts, b.Receipts()...)
+	}
+
+	// Apply end-of-block ops, mirroring the historical block loop above.
+	// The preliminary block is needed because EndOfBlockOps requires a
+	// *types.Block. It MUST include the same settlement receipts as the final
+	// block, otherwise block-sensitive hooks can derive different ops here and
+	// during execution.
+	preBlk, err := builder.BuildBlock(hdr, included, receipts)
```
JonathanOppenheimer: also, building the block didn't seem that expensive, to be able to use `EndOfBlockOps`

sae/blocks.go
```diff
@@ -241,8 +241,33 @@ func (vm *VM) buildBlock(
 		included = append(included, tx)
 	}
 
-	// TODO: Should the [hook.BlockBuilder] populate [types.Header.GasUsed] so
-	// that [hook.Op.Gas] can be included?
```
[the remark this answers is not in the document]
JonathanOppenheimer: Discussed on call the proper approach.

hook/hook.go
```diff
@@ -62,6 +62,14 @@ type Points interface {
 	AfterExecutingBlock(*state.StateDB, *types.Block, types.Receipts)
 }
 
+// State provides worst-case gas accounting for end-of-block operations.
+// [BlockBuilder.BuildBlock] implementations use it to apply [Op]s and query
+// cumulative gas.
+type State interface {
+	Apply(Op) error
+	GasUsed() uint64
```
JonathanOppenheimer: The hook needs this to populate the header... right....

PR #236, opened 2026-02-25
perf(saexec): Remove TrieDB memory leak

alarso16:
Currently, SAE follows this pattern:
If we call `triedb.Commit()` every `N` blocks:

State `0` will be on disk
States `[1, N]` will be generated in memory
State `N` will be committed, moving all dirty nodes at that root from memory onto disk
All outdated nodes in `[1, N)` will remain in the dirty cache

Although not a correctness bug, keeping these tries is generally unnecessary and leads to a memory leak.

The expected functionality should be:
- For validators: keep the minimum states around to guarantee that all consensus-necessary states are accessible
- For API nodes: keep all states on disk.

Now, all nodes will default to NOT storing all states. API nodes must set the necessary config

### DIFF
--- a/saedb/tracker.go
+++ b/saedb/tracker.go
@@ -73,8 +73,6 @@ func NewTracker(db ethdb.Database, c Config, lastExecuted common.Hash, log loggi
 //
 // This state will be available in memory until [Tracker.Untrack] has been
 // called for the root as many times as [Tracker.Track] has been called.
-//
-// Note: Snapshot memory leaks are avoided internally by [state.StateDB.Commit].
 func (t *Tracker) Track(root common.Hash, height uint64) error {
 	// Because [Tracker.Untrack] is always expected to be called (whether the state root changed or not),
 	// we must always add an additional reference


sae/recovery_test.go
```diff
@@ -67,8 +67,13 @@ func TestRecoverFromDatabase(t *testing.T) {
 			require.Len(t, b.Transactions(), 1, "transactions in block")
 		}
 		require.NoErrorf(t, b.WaitUntilExecuted(ctx), "%T.WaitUntilExecuted()", b)
+		t.Log(b.NumberU64(), b.PostExecutionStateRoot())
 
 		if quick {
+			t.Log("quick, skipping test")
+			if b.NumberU64() == 40 {
+				_ = b
+			}
```
JonathanOppenheimer: What is this for?

saedb/saedb.go
```diff
@@ -34,3 +47,129 @@ func LastCommittedTrieDBHeight(atOrBefore uint64) uint64 {
...
+	return state.New(root, e.cache, e.snaps)
+}
+
+// Record tracks the root and may commit the trie associated with the root
+// to the database if the height is on an multiple of [CommitTrieDBEvery].
+//
+// Note: Snapshot memory leaks are avoided internally by [state.StateDB.Commit].
+func (e *StateRecorder) Record(root common.Hash, height uint64) error {
+	// Push root if unique - don't want to remove an in-use state!
+	next, ok := e.inMemory.Index(e.inMemory.Len() - 1)
```
JonathanOppenheimer:
```suggestion
	last, ok := e.inMemory.Index(e.inMemory.Len() - 1)
```

saexec/saexec_test.go
```diff
@@ -893,3 +889,60 @@ func TestSnapshotPersistence(t *testing.T) {
...
+
+	final := chain.Last()
+	require.NoErrorf(t, final.WaitUntilExecuted(ctx), "%T.WaitUntilExecuted() on last-enqueued block", final)
+
+	t.Run("remove in memory state", func(t *testing.T) {
+		// We expect to not find blocks older than [saedb.StateHistory]
+		for _, b := range chain.AllBlocks() {
+			sdb, err := e.StateDB(b.PostExecutionStateRoot())
+			inMemory := b.NumberU64()+saedb.StateHistory > uint64(numBlocks) //nolint:gosec // positive plus positive
+			if saedb.ShouldCommitTrieDB(b.NumberU64()) || inMemory {
```
JonathanOppenheimer: this will only trigger for the genesis block because we have only 42 blocks and `CommitTrieDBEvery = 4096`. Do you want to either increase the block count or change `CommitTrieDBEvery` to make this test a little more thorough.

JonathanOppenheimer (changes requested):

saexec/saexec_test.go
```diff
@@ -893,3 +889,60 @@ func TestSnapshotPersistence(t *testing.T) {
...
+				require.IsTypef(t, &trie.MissingNodeError{}, err, "%T.StateDB() should not find block %d", e.StateRecorder, b.NumberU64())
+			}
+		}
+	})
+
+	t.Run("recover", func(t *testing.T) {
+		// Restart the chain to remove the TrieDB cache.
+		require.NoErrorf(t, sut.Close(), "%T.Close()", e)
+		src := blocks.Source(chain.GetBlock)
+		e, err := New(final, src.AsHeaderSource(), sut.chainConfig, sut.db, sut.xdb, &triedb.Config{}, defaultHooks(), sut.log) // TODO: share parameters better
```
JonathanOppenheimer: yeah you called this out, but bc of the log you're accessing internal executor state

sae/recovery_test.go
alarso16: Oops - debugging

saexec/saexec.go
```diff
@@ -28,11 +24,9 @@ import (
 	"github.com/ava-labs/strevm/saedb"
 )
 
-// SnapshotCacheSizeMB is the snapshot cache size used by the executor.
-const SnapshotCacheSizeMB = 128
-
 // An Executor accepts and executes a [blocks.Block] FIFO queue.
 type Executor struct {
+	*saedb.StateRecorder
```
ARR4N: This exposes `Record()`, which IIUC MUST only be called by the `Executor`.

saexec/saexec_test.go
```diff
@@ -92,17 +96,26 @@ func newSUT(tb testing.TB, hooks *saehookstest.Stub) (context.Context, SUT) {
...
 
-	return ctx, SUT{
+	sut := &SUT{
 		Executor: e,
 		chain:    chain,
 		wallet:   wallet,
 		logger:   logger,
 		db:       db,
 	}
+	tb.Cleanup(func() {
```
ARR4N: If you change the `closeOnce` field to type `func() error` then you can simplify this with `sync.OnceValue()`.

```go
	e, err := New(genesis, src.AsHeaderSource(), config, db, xdb, tdbConfig, hooks, logger)
	require.NoError(tb, err, "New()")
	closeOnce := sync.OnceValue(e.Close)
	tb.Cleanup(func() {
		require.NoErrorf(tb, closeOnce(), "%T.Close()", e)
	})

	return ctx, &SUT{
		Executor:  e,
		chain:     chain,
		wallet:    wallet,
		logger:    logger,
		db:        db,
		closeOnce: closeOnce,
	}
```

Note that it's `Executor` in the cleanup error because that's the type that you're closing, not the `SUT`.

saedb/saedb.go
```diff
@@ -34,3 +47,129 @@ func LastCommittedTrieDBHeight(atOrBefore uint64) uint64 {
...
+	if !ok {
+		return errors.New("no known state roots")
+	}
+	if next != root {
+		e.inMemory.Push(root)
+	}
```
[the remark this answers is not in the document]
alarso16: Arran asked about it too, so I guess I'll make a ~better~ comment

saexec/saexec_test.go
alarso16: Ah neat! I though about using this, but I worried the `OnceValue` hides the actual logic. But if you're not worried, then nice cleanup

saedb/saedb.go
```diff
@@ -34,3 +47,129 @@ func LastCommittedTrieDBHeight(atOrBefore uint64) uint64 {
...
+// Record tracks the root and may commit the trie associated with the root
+// to the database if the height is on an multiple of [CommitTrieDBEvery].
+//
+// Note: Snapshot memory leaks are avoided internally by [state.StateDB.Commit].
+func (e *StateRecorder) Record(root common.Hash, height uint64) error {
+	// Push root if unique - don't want to remove an in-use state!
+	next, ok := e.inMemory.Index(e.inMemory.Len() - 1)
+	if !ok {
+		return errors.New("no known state roots")
+	}
```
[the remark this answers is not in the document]
alarso16: Yeah it's logically impossible. I'll collapse it. Then I don't have to push the first block to avoid this edge case

saedb/saedb.go
```diff
@@ -34,3 +47,129 @@ func LastCommittedTrieDBHeight(atOrBefore uint64) uint64 {
...
+	StateHistory = 32
+)
+
+// NewStateRecorder returns a [StateRecorder] object that abstracts database interactions on the EVM state.
+//
+// TODO(alarso16): Provide a custom config to generate the [triedb.Config].
+func NewStateRecorder(db ethdb.Database, c *triedb.Config, lastExecuted common.Hash, log logging.Logger) (*StateRecorder, error) {
+	cache := state.NewDatabaseWithConfig(db, c)
+	q, err := buffer.NewBoundedQueue(StateHistory, func(root common.Hash) {
+		// This error doesn't need to be fatal, since all current state is likely still valid.
```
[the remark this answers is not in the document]
alarso16: Comment removed - the error is interesting...

saexec/saexec_test.go
JonathanOppenheimer: ah awesome!

sae/consensus.go
```diff
@@ -109,9 +109,11 @@ func (vm *VM) AcceptBlock(ctx context.Context, b *blocks.Block) error {
 			continue
 		}
 		vm.blocks.Delete(s.Hash())
+		vm.exec.Untrack(s.PostExecutionStateRoot())
```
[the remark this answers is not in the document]
ARR4N: I think the second half of my block is incorrect and should follow the same pattern as in the loop.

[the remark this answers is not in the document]
alarso16: Each time a state root is seen by the Tracker, a reference is added. Thus, it won't removed until Untrack is called as many times as Track was called. Since each block will call Track, even if the state root is not unique, everything works fine

saedb/tracker.go
```diff
@@ -0,0 +1,158 @@
...
+type Tracker struct {
+	snaps       *snapshot.Tree
+	cache       state.Database
+	isHashDB    bool
+	isArchival  bool
+	log         logging.Logger
+	currentRoot common.Hash
+}
+
+// NewTracker provides a new [Tracker] on the underlying database
```
ARR4N:
```suggestion
// NewTracker provides a new [Tracker] on the underlying database.
```

(nit) https://google.github.io/styleguide/go/decisions#comment-sentences

saexec/saexec.go
```diff
@@ -74,25 +61,21 @@ func New(
...
 	if err != nil {
 		return nil, err
 	}
 
 	e := &Executor{
-		quit:  make(chan struct{}), // closed by [Executor.Close]
-		done:  make(chan struct{}), // closed by [Executor.processQueue] after `quit` is closed
-		log:   log,
-		hooks: hooks,
+		Tracker: s,
```
ARR4N:
```suggestion
		Tracker: t,
```

See above. Bundle with this.

PR #243, opened 2026-02-25
fix: use maxFutureBlockTime

### LINKED ISSUE #242: Bug: Incorrect block timestamp checks
We currently have two constants in the code that restrict timestamps on blocks:
1. https://github.com/ava-labs/strevm/blob/f92d4f6ac0c252caa0fcbb723a94721ee4135c98/sae/blocks.go#L42
2. https://github.com/ava-labs/strevm/blob/f92d4f6ac0c252caa0fcbb723a94721ee4135c98/sae/blocks.go#L77-L79

`maxBlockFutureSeconds` was intended to be enforced during `ParseBlock` to avoid overflow cases when converting the block's timestamp to a `time.Time`.

`maxFutureBlockTime` was intended to be enforced during `VerifyBlock` to prevent the chain from advancing too far into the future.

---

`maxFutureBlockTime` is currently only used in tests. In `VerifyBlock` we incorrectly use `maxBlockFutureSeconds` as a duration.

JonathanOppenheimer:
Closes #242. 

There were two separate constants for the allowable future time. We consolidate to one, update the tests appropriately and add a new test for blocks on the boundary.

### DIFF
--- a/sae/vm_test.go
+++ b/sae/vm_test.go
@@ -612,14 +612,14 @@ func TestSyntacticBlockChecks(t *testing.T) {
 			name: "block_time_at_maximum",
 			header: &types.Header{
 				Number: big.NewInt(1),
-				Time:   now + uint64(maxFutureBlockTime.Seconds()),
+				Time:   now + maxFutureBlockSeconds,
 			},
 		},
 		{
 			name: "block_time_after_maximum",
 			header: &types.Header{
 				Number: big.NewInt(1),
-				Time:   now + uint64(maxFutureBlockTime.Seconds()) + 1,
+				Time:   now + maxFutureBlockSeconds + 1,
 			},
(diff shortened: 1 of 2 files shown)


sae/vm_test.go
```diff
@@ -717,11 +726,6 @@ func TestSemanticBlockChecks(t *testing.T) {
 			time:    lastAccepted.BuildTime() - 1,
 			wantErr: errBlockTimeBeforeParent,
 		},
-		{
-			name:    "block_time_after_maximum",
-			time:    now + uint64(maxFutureBlockTime.Seconds()) + 1,
-			wantErr: errBlockTimeAfterMaximum,
-		},
```
StephenButtolph: It feels bad deleting this test... How can we test the if statement in `VerifyBlock`?

PR #244, opened 2026-02-26
feat: Include `hook.Op`s during block building

### LINKED ISSUE #193: Should `hook.BlockBuilder` populate `GasUsed`?

StephenButtolph:
This PR adds operation support to block building.

`hook.BlockBuilder` is made generic and it is moved from `hook.Points` to a new interface, `hook.PointsG`. `SinceGenesis` and `NewVM` are also made generic, but the `VM` struct itself is not made generic to avoid polluting the generic type everywhere. 

`hooktest.Stub` is extended to support encoding `Op`s in the blocks that it generates.

Finally, an integration test is introduced that verifies that ops are correctly included into blocks and executed.

Closes #193

### DIFF
--- a/blocks/invariants.go
+++ b/blocks/invariants.go
@@ -16,6 +16,7 @@ import (
 	"github.com/holiman/uint256"
 	"go.uber.org/zap"
 
+	"github.com/ava-labs/strevm/gastime"
 	"github.com/ava-labs/strevm/hook"
 )
 
@@ -23,6 +24,11 @@ import (
 // builder, that a [Block] will encounter when eventually executed.
 type WorstCaseBounds struct {
 	MaxBaseFee *uint256.Int
+	// LatestEndTime is the worst-case [gastime.Time] after this block's gas has been
+	// consumed and the target updated. Its [gastime.Time.BaseFee] is an upper
+	// bound on the next block's base fee because the next block's
+	// [gastime.Time.BeforeBlock] can only reduce the excess.
+	LatestEndTime *gastime.Time
 	// Invariant: keys of individual maps MUST be identical to those of the
 	// respective [hook.Op.Burn] map. For transaction-derived Ops, there is
 	// always 1 entry.
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -18,9 +18,7 @@ jobs:
   # what gates PRs.
   go:
     runs-on: ubuntu-latest
-    needs: [go_build_prod, go_test, go_test_race, go_generate, go_tidy, require_fuzz_corpus]
-    # TODO(arr4n) investigate why setup-go wasn't properly caching fuzz corpora
-    # and then reinstate a go_fuzz job that extends them.
+    needs: [go_build_prod, go_test, go_test_race, go_generate, go_tidy, go_fuzz]
     steps:
       - run: echo "Dependencies successful"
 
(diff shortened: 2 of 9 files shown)


blocks/invariants.go
```diff
@@ -109,6 +109,13 @@ func (b *Block) CheckOpBurnerBalanceBounds(stateDB *state.StateDB, opIndexInBloc
 }
 
 func (b *Block) checkBalanceBounds(log logging.Logger, stateDB *state.StateDB, opIndexInBlock int, accounts ...common.Address) {
+	if n := len(b.bounds.MinOpBurnerBalances); opIndexInBlock >= n {
+		log.Warn("Op not included in worst-case bounds",
+			zap.Int("num_ops", n),
+		)
+		return
+	}
+
```
StephenButtolph: Unrelated to the rest of the PR, but included to address the ToB issue in a single PR/commit.

hook/hookstest/stub.go
```diff
@@ -94,15 +133,26 @@ func (s *Stub) GasTargetAfter(*types.Header) gas.Gas {
 // [Stub.BuildHeader] in the header's `Extra` field. If said field is empty,
 // SubSecondBlockTime returns 0.
 func (s *Stub) SubSecondBlockTime(hdr *types.Header) time.Duration {
-	if len(hdr.Extra) == 0 {
-		return 0
+	e := extra{}
+	if err := e.UnmarshalCanoto(hdr.Extra); err != nil {
+		// This is left as a panic to avoid polluting various functions with
+		// error returns when no error is possible in production.
+		panic(err)
```
JonathanOppenheimer: Can you explain the rationale behind this -- this is a test, and you checked the same error in `EndOfBlockOps`, `BlockRebuilderFrom`.

hook/hookstest/stub.go
```diff
@@ -121,3 +171,63 @@ func (*Stub) BeforeExecutingBlock(params.Rules, *state.StateDB, *types.Block) er
...
+	Gas       gas.Gas         `canoto:"uint,2"`
+	GasFeeCap uint256.Int     `canoto:"fixed repeated uint,3"`
+	Burn      []AccountDebit  `canoto:"repeated value,4"`
+	Mint      []AccountCredit `canoto:"repeated value,5"`
+
+	canotoData canotoData_Op
+}
+
+// AsOp converts the op into a representation that SAE can use directly.
+func (o Op) AsOp() hook.Op {
```
JonathanOppenheimer: (question) -- it's not possible for there to be duplicates of addresses in `Burn` or `Mint` right?

sae/vm_test.go
```diff
@@ -656,6 +656,95 @@ func TestCanCreateContractSoftError(t *testing.T) {
...
+		},
+		{
+			Gas:       200_000,
+			GasFeeCap: gasFeeCap,
+			Mint: []hookstest.AccountCredit{
+				{
+					Address: receiver,
+					Amount:  *uint256.NewInt(received),
+				},
+			},
```
JonathanOppenheimer: Would a case for checking the error on `BlockRebuilderFrom` (more coverage) make sense?

hook/hookstest/stub.go
StephenButtolph: This is a fair question.

I added errors to functions that I felt could encounter errors in production.

`SubSecondBlockTime` is going to be implemented with a libevm hook on the header. So I don't think coreth or subnet-evm will actually ever encounter an error in this function.

Additionally, I wanted to avoid adding errors to functions that would cause a lot of error bubbling. For this function, we currently call it in a number of functions that currently don't return errors. So returning an error here would add a lot of error checks.

`EndOfBlockOps` and `BlockRebuilderFrom` are both probably going to be parsing atomic txs in coreth, which can error.

hook/hookstest/stub.go
StephenButtolph: `hookstest.Op` does not prevent duplicates from existing, it ignores the duplicates and just returns the last address instances.

In the production code, `hook.Op` doesn't allow duplicates, since it is a map.

saexec/execution.go
```diff
@@ -51,6 +51,8 @@ func (e *Executor) Enqueue(ctx context.Context, block *blocks.Block) error {
 	}
 }
 
+var emergencyPlaybookLink = zap.String("playbook", "https://github.com/ava-labs/strevm/issues/28")
```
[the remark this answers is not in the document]
ARR4N: We're now going to get double logging of this when done both inside and on the return of `execute()`, and also once for non-emergency errors like committing state. I think it should only be done in `processQueue()` and i.f.f. the returned error wraps some `errEmergency` sentinel (or as a fail closed, not if the returned error wraps some `errNonEmergency` sentinel).

sae/block_builder.go
```diff
@@ -97,12 +107,16 @@ func (b *blockBuilder) rebuild(
 		}
 	}
 
+	rebuilder, err := b.hooks.BlockRebuilderFrom(block.EthBlock())
+	if err != nil {
+		return nil, fmt.Errorf("making rebuilder: %v", err)
```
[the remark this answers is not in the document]
StephenButtolph: fwiw, the consensus code includes the blockID and height (and who gave us the block) in the log where this is reported. But because the proposervm modified the blockID when adding it's header, I agree that including the SAE block hash improves debug-ability.

hook/hookstest/stub.go
```diff
@@ -54,35 +58,70 @@ func (s *Stub) BuildHeader(parent *types.Header) *types.Header {
...
+
+// BuildBlock encodes ops into [types.Header.Extra] and calls [types.NewBlock]
+// with the other arguments.
+func BuildBlock(
+	header *types.Header,
+	txs []*types.Transaction,
+	receipts []*types.Receipt,
+	ops []Op,
 ) (*types.Block, error) {
+	e := extra{}
```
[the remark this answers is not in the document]
StephenButtolph: lol, I guess we prefer to `var` rather than `:=` here.

I updated to use `var` since it saves us exactly one character of code per line. Gotta help with the disk shortage any way we can. 😜

hook/hook.go
```diff
@@ -89,9 +102,16 @@ type BlockBuilder interface {
 		header *types.Header,
 		txs []*types.Transaction,
 		receipts []*types.Receipt,
+		endOfBlockOps []T,
 	) (*types.Block, error)
 }
 
+// Transaction is a user defined transaction type that can be represented as an
```
```suggestion
// Transaction is a user-defined transaction type that can be represented as an
```

(nit)

ARR4N (approved):
No need for a re-review of changes.

PR #245, opened 2026-02-26
feat(sae): implement `StateAtBlock` and `StateAtTransaction` for state tracing support

powerslider:
- Implement `apiBackend.StateAtBlock()` and `apiBackend.StateAtTransaction()` to fully satisfy the `tracers.Backend` interface, enabling `debug_traceTransaction`, `debug_traceBlockByNumber`, and related tracing RPCs.
- Resolve post-execution state roots and base fees from memory (`vm.blocks.Load`) or disk (`blocks.PostExecutionStateRoot`, `blocks.ExecutionBaseFee`), bypassing geth's re-execution-from-archive strategy.
- Extract a standalone `saexec.Execute()` function from `Executor.execute()` to guarantee identical transaction application ordering between the executor and the tracer, eliminating divergence risk. Both callers share the same code path - the only differences are injected via parameters:
  - `maxNumTxs` - executor passes all transactions, tracer passes `txIndex` for partial replay.
  - `hooks` - tracer wraps hooks with `noEndOfBlockOps` to suppress end-of-block operations.
  - `receiptStore` - tracer passes `NullReceiptStore` to skip receipt broadcasting.
- Move `StateDBOpener` interface to `saedb` package for shared use by executor and tracer.
- `StateAtTransaction` constructs fresh `*blocks.Block` instances via `blockBuilder.new()` + `RestoreExecutionArtefacts()` to avoid passing live in-memory blocks to `Execute()` concurrently with the executor.
- Add `TestTraceContractInteraction` covering `debug_traceTransaction` and `debug_traceBlockByNumber` through the RPC API, verifying EVM-level trace correctness by tracing escrow contract deployment and deposit, asserting `SSTORE`/`SLOAD` opcodes.

### DIFF
--- a/sae/rpc_stateful.go
+++ b/sae/rpc_stateful.go
@@ -96,7 +96,7 @@ func (b *apiBackend) StateAndHeaderByNumberOrHash(ctx context.Context, numOrHash
 		hdr.Root = bl.PostExecutionStateRoot()
 		hdr.BaseFee = bl.BaseFee().ToBig()
 	} else {
-		hdr = rawdb.ReadHeader(b.db, hash, num)
+		hdr = rawdb.ReadHeader(b.vm.db, hash, num)
 
 		// TODO(arr4n) export [blocks.executionResults] to avoid multiple
 		// database reads and canoto unmarshallings here.
@@ -113,7 +113,7 @@ func (b *apiBackend) StateAndHeaderByNumberOrHash(ctx context.Context, numOrHash
 		hdr.BaseFee = bf.ToBig()
 	}
 
-	sdb, err := state.New(hdr.Root, b.exec.StateCache(), nil)
+	sdb, err := state.New(hdr.Root, b.vm.exec.StateCache(), nil)
 	if err != nil {
 		return nil, nil, err
 	}
--- a/blocks/blockstest/chain.go
+++ b/blocks/blockstest/chain.go
@@ -11,7 +11,6 @@ import (
 	"testing"
 
 	"github.com/ava-labs/libevm/common"
-	"github.com/ava-labs/libevm/core"
 	"github.com/ava-labs/libevm/core/types"
 	"github.com/ava-labs/libevm/event"
 	"github.com/ava-labs/libevm/libevm/options"
@@ -25,10 +24,10 @@ import (
 //
 // It is not safe for concurrent use.
 type ChainBuilder struct {
-	config       *params.ChainConfig
-	chain        []*blocks.Block
-	blocksByHash sync.Map
-	headEvents   event.FeedOf[core.ChainHeadEvent]
+	config         *params.ChainConfig
+	chain          []*blocks.Block
+	blocksByHash   sync.Map
+	acceptedBlocks event.FeedOf[*blocks.Block]
 
 	defaultOpts []ChainOption
 }
@@ -88,7 +87,7 @@ func (cb *ChainBuilder) NewBlock(tb testing.TB, txs []*types.Transaction, opts .
 
 	cb.chain = append(cb.chain, b)
 	cb.blocksByHash.Store(b.Hash(), b)
-	cb.headEvents.Send(core.ChainHeadEvent{Block: b.EthBlock()})
+	cb.acceptedBlocks.Send(b)
 
 	return b
 }
@@ -129,10 +128,10 @@ func (cb *ChainBuilder) GetBlock(h common.Hash, num uint64) (*blocks.Block, bool
 // chain height.
 var ErrBlockNotFound = errors.New("block not found")
 
-// SubscribeChainHeadEvent subscribes to chain head events fired by
+// SubscribeAcceptedBlocks subscribes to accepted block events fired by
 // [ChainBuilder.NewBlock].
-func (cb *ChainBuilder) SubscribeChainHeadEvent(ch chan<- core.ChainHeadEvent) event.Subscription {
-	return cb.headEvents.Subscribe(ch)
+func (cb *ChainBuilder) SubscribeAcceptedBlocks(ch chan<- *blocks.Block) event.Subscription {
+	return cb.acceptedBlocks.Subscribe(ch)
 }
 
 // LastAcceptedBlock returns the last block in the chain.
--- a/gasprice/estimator.go
+++ b/gasprice/estimator.go
@@ -17,11 +17,11 @@ import (
 
 	"github.com/ava-labs/avalanchego/utils/logging"
 	"github.com/ava-labs/libevm/common/math"
-	"github.com/ava-labs/libevm/core"
 	"github.com/ava-labs/libevm/core/types"
 	"github.com/ava-labs/libevm/event"
 	"github.com/ava-labs/libevm/params"
 	"github.com/ava-labs/libevm/rpc"
+	"go.uber.org/zap"
 
 	"github.com/ava-labs/strevm/blocks"
 	"github.com/ava-labs/strevm/intmath"
@@ -31,7 +31,7 @@ import (
 type Backend interface {
 	ResolveBlockNumber(bn rpc.BlockNumber) (uint64, error)
 	BlockByNumber(bn rpc.BlockNumber) (*types.Block, error)
-	SubscribeChainHeadEvent(ch chan<- core.ChainHeadEvent) event.Subscription
+	SubscribeAcceptedBlocks(ch chan<- *blocks.Block) event.Subscription
 	LastAcceptedBlock() *blocks.Block
 }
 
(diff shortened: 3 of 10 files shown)


sae/rpc_stateful_test.go
```diff
@@ -8,21 +8,138 @@ import (
...
+	settled := sut.runConsensusLoop(t, createTx(t, recv), createTx(t, recv))
+	vmTime.advanceToSettle(ctx, t, settled)
+
+	// Build after settlement so it stays in memory.
+	executed := sut.runConsensusLoop(t, createTx(t, recv), createTx(t, recv))
+	require.NoErrorf(t, executed.WaitUntilExecuted(ctx), "%T.WaitUntilExecuted()", executed)
+
+	pending := sut.runConsensusLoop(t, createTx(t, blockingPrecompile))
+
+	be := sut.rawVM.apiBackend
```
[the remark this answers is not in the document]
powerslider: Check https://github.com/ava-labs/strevm/pull/245/changes/672b4664aedf832dd51111e74723321ef170e1b9

sae/rpc_stateful.go
```diff
@@ -104,3 +107,123 @@ func (b *ethAPIBackend) StateAndHeaderByNumberOrHash(ctx context.Context, numOrH
...
+	// hooks run before the header is modified with the execution base fee.
+	chainConfig := b.ChainConfig()
+	rules := chainConfig.Rules(block.Number(), true /*isMerge*/, block.Time())
+	if err := b.vm.exec.Hooks().BeforeExecutingBlock(rules, stateDB, block); err != nil {
+		return nil, vm.BlockContext{}, nil, nil, err
+	}
+
+	// Replace the worst-case consensus base fee in the header with the
+	// actual execution base fee, as done during block execution.
+	header := types.CopyHeader(block.Header())
```
alarso16: You're double-copying that header

sae/rpc_stateful.go
```diff
@@ -104,3 +107,123 @@ func (b *ethAPIBackend) StateAndHeaderByNumberOrHash(ctx context.Context, numOrH
...
+	}
+
+	signer := types.MakeSigner(chainConfig, block.Number(), block.Time())
+	blockCtx := core.NewEVMBlockContext(header, b.vm.exec.ChainContext(), &header.Coinbase)
+	gasPool := core.GasPool(math.MaxUint64)
+
+	// Replay transactions 0..txIndex-1 to produce the state just before the
+	// target transaction.
+	for i, tx := range txs[:txIndex] {
+		stateDB.SetTxContext(tx.Hash(), i)
```
alarso16: This line took me a while to convince myself that it's necessary, but the logs are never tested (which is what this is used for)

sae/rpc_stateful_test.go
```diff
@@ -4,25 +4,250 @@
...
+		require.Equalf(t, types.ReceiptStatusSuccessful, r.Status, "%T.Status", r)
+	}
+
+	// Settle and evict so the block is only accessible from disk.
+	vmTime.advanceToSettle(ctx, t, b)
+	for range 2 {
+		bb := sut.runConsensusLoop(t)
+		vmTime.advanceToSettle(ctx, t, bb)
+	}
+	hasOp := func(logs []structLogEntry, op string) bool {
```
JonathanOppenheimer: Do you want to move this to before the settlement block, or make this just a top level function for clarity?

sae/rpc_stateful_test.go
```diff
@@ -4,25 +4,250 @@
...
+				return true
+			}
+		}
+		return false
+	}
+
+	// Trace the deposit tx (txIndex=1). The tracer must replay the deploy
+	// tx first via StateAtTransaction, then trace the deposit which writes
+	// to contract storage (SSTORE) and emits the Deposit event (LOG1).
+	t.Run("debug_traceTransaction", func(t *testing.T) {
```
JonathanOppenheimer: Do you want to add a test for calling `debug_traceTransaction` with a tx hash that doesn't exist?

JonathanOppenheimer (commented):
Agree w/ a lot of Austin's comments, a few additional things.

saexec/execution.go
```diff
@@ -12,111 +12,105 @@ import (
...
-				zap.Uint64("block_time", block.BuildTime()),
-				zap.Stringer("block_hash", block.Hash()),
-				zap.Int("tx_count", len(block.Transactions())),
-			)
+	// NoEndOfBlockOps wraps [hook.Points] to suppress
+	// [hook.Points.EndOfBlockOps], used by the tracer to skip end-of-block
+	// operations during partial replay.
+	NoEndOfBlockOps struct {
+		hook.Points
+	}
```
ARR4N: This doesn't belong here if it's only for the tracer. As a rule of thumb, if a comment in package `X` says that something is for a type in package `Y` then it's suggestive that it might be misplaced.

saexec/execution.go
```diff
@@ -12,111 +12,105 @@ import (
...
+	signer := types.MakeSigner(config, ethBlock.Number(), ethBlock.Time())
+	gasPool := core.GasPool(math.MaxUint64)
 	var blockGasConsumed gas.Gas
 
-	signer := e.SignerForBlock(b.EthBlock())
-	receipts := make(types.Receipts, len(b.Transactions()))
-	for ti, tx := range b.Transactions() {
+	txs := b.Transactions()
+	txs = txs[:min(len(txs), maxNumTxs)]
+	allReceipts := make(types.Receipts, len(txs))
```
ARR4N: Why is this variable name changed?

saexec/execution.go
```diff
@@ -225,9 +308,19 @@ func (e *Executor) execute(b *blocks.Block, logger logging.Logger) error {
...
+	if err := b.MarkExecuted(e.db, e.xdb, r.FinishBy.Gas.Clone(), r.FinishBy.Wall, r.Header.BaseFee, r.Receipts, root, &e.lastExecuted /* (2) */); err != nil {
 		return err
 	}
-	e.sendPostExecutionEvents(b.EthBlock(), receipts) // (3)
+	e.sendPostExecutionEvents(b.EthBlock(), r.Receipts) // (3)
 	return nil
 }
+
+// NullReceiptStore is a no-op [ReceiptStore] for use when receipt broadcasting
+// is not needed (e.g. state tracing).
```
ARR4N: (fyi, no action required) I think providing an example of usage is acceptable in a comment because it's demonstrative of how one might (correctly) use the type. I appreciate that I'm being very picky about the comment wording but precision is important because doc comments are occasionally the only thing a consumer sees.

saexec/saexec.go
```diff
@@ -149,6 +149,11 @@ func (e *Executor) ChainContext() core.ChainContext {
 	return e.chainContext
 }
 
+// StateDB implements [StateDBOpener].
```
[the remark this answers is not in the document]
powerslider: Absolutely and this is my bad. I always add compile time interface checks. Sometimes when I am lazy I put a comment to remind myself to add it, but if I forgot in this case.

sae/rpc_stateful.go
```diff
@@ -104,3 +123,120 @@ func (b *apiBackend) StateAndHeaderByNumberOrHash(ctx context.Context, numOrHash
...
+func (vm *VM) executionBaseFee(hash common.Hash, num uint64) (*big.Int, error) {
+	if bl, ok := vm.blocks.Load(hash); ok {
+		return bl.BaseFee().ToBig(), nil
+	}
+	bf, err := blocks.ExecutionBaseFee(vm.xdb, num)
+	if err != nil {
+		return nil, err
+	}
+	return bf.ToBig(), nil
+}
```
ARR4N:
```suggestion
```

This is only used once, in a place that needs to be reverted.

sae/rpc_stateful.go
```diff
@@ -104,3 +123,120 @@ func (b *apiBackend) StateAndHeaderByNumberOrHash(ctx context.Context, numOrHash
...
+		&saexec.NullReceiptStore{},
+		b.vm.log(),
+	)
+	if err != nil {
+		return nil, bCtx, nil, nil, err
+	}
+
+	// Prepare the target transaction's message.
+	targetTx := txs[txIndex]
+	msg, err := core.TransactionToMessage(targetTx, result.Signer, result.BaseFee.ToBig())
```
ARR4N:
```suggestion
	msg, err := core.TransactionToMessage(txs[txIndex], result.Signer, result.BaseFee.ToBig())
```

(readability) The `targetTx` variable is only used on two lines and doesn't provide the reader with any information so is redundant, but does signal "you need to remember this". As above re "what" comments.

Issue #74, opened 2026-01-12
`worstcase.txToOp()` over-charges dynamic base fees


ARR4N:
`types.Transaction.Cost()` uses `GasPrice()`, which is equivalent to `GasFeeCap()`. However we know the worst-case base fee and should be using that instead. This probably isn't an urgent issue because the mempool wouldn't allow said tx in anyway, but it's still not a correct implementation.

PR #249, opened 2026-02-26
fix: worst-case state charges effective gas price instead of fee cap

JonathanOppenheimer:
Closes https://github.com/ava-labs/strevm/issues/74. 

Previously, `worstcase.txToOp()` used `tx.Cost()` (which uses `GasFeeCap` as the price per gas) to both validate a transaction for inclusion and debit the account. In geth's `buyGas`, the balance is checked against `GasLimit * GasFeeCap + Value`, but only `GasLimit * EffectiveGasPrice + Value` is actually debited.                                              

While this previous approach was perfectly safe, it was overly conservative -- it drained account balances faster than the actual EVM would, which could cause it to reject later transactions from the same sender that would have been perfectly valid during execution. @StephenButtolph ran into this while using the dev-net, so changing this is a meaningful UX improvement. Don't tell someone they don't have enough money to do something when they really do have enough money!


hook/hook_test.go
```diff
@@ -102,11 +111,55 @@ func TestOp_ApplyTo(t *testing.T) {
...
+				},
+			},
+			wantAccounts: []account{
+				{
+					address: eoa,
+					nonce:   3,
+					balance: uint256.NewInt(400),
+				},
+			},
+		},
```
JonathanOppenheimer: This test is a "narrative" test that @ARR4N has been against. We test mint -> transfer -> burn -> overdraft and then I added a fund -> too expensive -> transfer -> validate balance.

I am particularly not a fan of the `fund_eoa_for_max_amount_tests` step, but I also do not want to rewrite this test wholesale.

hook/hook.go
```diff
@@ -91,10 +93,26 @@ type BlockBuilder interface {
...
 type AccountDebit struct {
-	Nonce  uint64
+	Nonce uint64
+	// Amount is the effective cost to deduct from the account balance.
 	Amount uint256.Int
+	// MaxAmount is the maximum cost the account must be able to afford.
+	// This mirrors geth's balanceCheck in buyGas: the account is validated
+	// against MaxAmount (e.g. gasLimit * gasFeeCap + value) but only
+	// charged Amount (e.g. gasLimit * effectiveGasPrice + value).
+	MaxAmount uint256.Int
```
StephenButtolph: `MaxAmount` doesn't make too much sense to me (since we are specifying the specific amount). Can we instead call it `MinBalance`?
```suggestion
	// Amount to deduct from the account balance.
	Amount uint256.Int
	// MinBalance is the minimum balance the account must have for the operation
	// to be valid. It must be at least [AccountDebit.Amount].
	MinBalance uint256.Int
```

hook/hook.go
```diff
@@ -117,11 +135,15 @@ type Op struct {
 
 // ApplyTo applies the operation to the statedb.
 //
-// If an account has insufficient funds, [core.ErrInsufficientFunds] is returned
-// and the statedb is unchanged.
+// If an account has insufficient funds, [core.ErrInsufficientFunds] is
+// returned. If an entry has malformed fields, [errInvalidAccountDebit] is returned.
```
StephenButtolph: I'm not sure if it makes sense to document an unexported error on an exported function. As people won't actually be able to inspect that behavior.

worstcase/state.go
```diff
@@ -216,31 +216,45 @@ func (s *State) ApplyTx(tx *types.Transaction) error {
...
+	var maxAmount uint256.Int
+	if overflow := maxAmount.SetFromBig(tx.Cost()); overflow {
+		return Op{}, errCostOverflow
+	}
+
+	// amount = gasLimit * effectiveGasPrice + value (actual deduction)
+	baseFeeBI := baseFee.ToBig()
+	effectiveGasPrice := new(big.Int).Add(baseFeeBI, tx.EffectiveGasTipValue(baseFeeBI))
+	cost := new(big.Int).Mul(new(big.Int).SetUint64(tx.Gas()), effectiveGasPrice)
+	cost.Add(cost, tx.Value())
```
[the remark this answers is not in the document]
JonathanOppenheimer: Let me know what you think of this.

hook/hook.go
```diff
@@ -91,12 +93,18 @@ type BlockBuilder interface {
...
 // AccountDebit includes an amount that an account should have debited,
-// along with the nonce used to aut debit the account.
+// along with the nonce used to authorize the debit.
 type AccountDebit struct {
-	Nonce  uint64
+	Nonce uint64
+	// Amount to deduct from the account balance.
 	Amount uint256.Int
+	// MinBalance is the minimum balance the account must have for the operation
+	// to be valid. It must be at least [AccountDebit.Amount].
```
```suggestion
	// to be valid. It MUST be at least [AccountDebit.Amount].
```

(nit) This is being used akin to RFC 2119 so it's idiomatic to capitalise it.

PR #250, opened 2026-02-26
refactor: Move block building off the VM

StephenButtolph:
This moves block building and rebuilding off of the `VM` onto a new `blockBuilder` struct. This is a (almost) non-functional refactor. It is intended to improve the readability of the later `hook.BlockBuilder` refactor PR (which will include functional changes to the block building logic).

### DIFF
--- a/sae/block_builder.go
+++ b/sae/block_builder.go
@@ -34,18 +34,18 @@ type blockBuilder struct {
 	mempool *txgossip.Set
 }
 
-func (b *blockBuilder) New(eth *types.Block, parent, lastSettled *blocks.Block) (*blocks.Block, error) {
+func (b *blockBuilder) new(eth *types.Block, parent, lastSettled *blocks.Block) (*blocks.Block, error) {
 	return blocks.New(eth, parent, lastSettled, b.log)
 }
 
-// Build a new block on top of the provided parent. The block context MAY be
+// build a new block on top of the provided parent. The block context MAY be
 // nil.
-func (b *blockBuilder) Build(
+func (b *blockBuilder) build(
 	ctx context.Context,
 	bCtx *block.Context,
 	parent *blocks.Block,
 ) (*blocks.Block, error) {
-	return b.build(
+	return b.buildWithTxs(
 		ctx,
 		bCtx,
 		parent,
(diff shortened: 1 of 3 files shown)


sae/blocks.go
```diff
@@ -287,14 +71,6 @@ var (
 // VerifyBlock validates the block and, if successful, populates its ancestry.
 // The block context MAY be nil.
 func (vm *VM) VerifyBlock(ctx context.Context, bCtx *block.Context, b *blocks.Block) error {
-	// Sender caching should be performed as early as possible but not at the
-	// risk of spam. [VM.ParseBlock] is too early as there is no protection,
-	// whereas [VM.VerifyBlock] is only called after verifying the current
-	// proposer's signature. While a malicious proposer could exist, their time
-	// window is limited.
```
StephenButtolph: I significantly reworded this comment, as it is now at essentially the last possible part of the code that it can be.

sae/vm_test.go
```diff
@@ -717,12 +717,10 @@ func TestAcceptBlock(t *testing.T) {
 
 func TestSemanticBlockChecks(t *testing.T) {
 	const now = 1e6
+	opt, _ := withVMTime(t, time.Unix(now, 0))
 	ctx, sut := newSUT(t, 1, options.Func[sutConfig](func(c *sutConfig) {
 		c.genesis.Timestamp = now
-	}))
-	sut.rawVM.config.Now = func() time.Time {
-		return time.Unix(now, 0)
-	}
```
StephenButtolph: This diff is required because the block builder is no longer referencing the `now` in the `rawVM`. It is passed during `NewVM` explicitly.

ARR4N: (low priority) When I saw the `blockBuilder` I thought it was carrying a lot of duplicate stuff. Should we be addressing it there instead of here?

Perhaps a broader refactor will be required, to add a carrier type that holds things like `snowCtx`, hooks, `Config`, etc.

ARR4N (approved):
No need for a re-review should you choose to heed my suggestions.

sae/vm_test.go
StephenButtolph: imo, this test should have been using `withVMTime` previously. So I don't think that needing to change this test really implies that something else needs to be changed in the VM implementation.

That aside, I'm not super against making a helper type for all the "configurations" of the VM... I generally try to avoid making union types so that functions can be more easily re-used/tested... But 🤷

sae/block_builder.go
```diff
@@ -0,0 +1,305 @@
...
+
+type blockBuilder struct {
+	hooks   hook.Points
+	now     func() time.Time
+	log     logging.Logger
+	exec    *saexec.Executor
+	mempool *txgossip.Set
+}
+
+func (b *blockBuilder) New(eth *types.Block, parent, lastSettled *blocks.Block) (*blocks.Block, error) {
```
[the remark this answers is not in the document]
ARR4N: As I said, very subjective, but I think it's better just because it's consistent in intent. Feel free to change them all at once in the next PR.

PR #252, opened 2026-02-27
chore: max transaction size explicit cap comment 

### LINKED ISSUE #187: MaxSize set to `math.MaxUint` in validation options
See: https://github.com/ava-labs/strevm/blob/5faf8141974efdc39dca07b179f74c18db7a8718/worstcase/state.go#L195

Transaction size limit is effectively disabled during worst-case validation.

Should probably be set to something real (e.g. match the network's actual max tx size) to avoid accepting oversized transactions during worst-case checks.

JonathanOppenheimer:
Closes #187 

Currently, the max size of an accepted transaction is explicitly uncapped -- this is okay because it implicitly is. This PR removes the TODO and explains why an explicit cap is unnecessary

### DIFF
--- a/worstcase/state.go
+++ b/worstcase/state.go
@@ -191,9 +191,9 @@ func (s *State) ApplyTx(tx *types.Transaction) error {
 			1<<types.LegacyTxType |
 			1<<types.AccessListTxType |
 			1<<types.DynamicFeeTxType,
-		// No byte-size limit needed: gas validation (intrinsic gas ≤
+		// No byte-size limit needed as gas validation (intrinsic gas ≤
 		// tx gas ≤ block gas limit) already enforces an implicit size
-		// bound on transactions.
+		// limit (2MB) on transactions.
 		MaxSize: math.MaxUint,
 		MinTip:  big.NewInt(0),
 	}


alarso16 (approved):
this seems straightforward

ARR4N:
> > There is already an implicit limit imposed by the block size and the charge for zero bytes. Why should that be reduced further? Keep in mind that a block like this is already subject to p2p and ProposerVM limitations so we're only considering an instance in which a builder wants to propose a block with a tx that someone is willing (and able) to pay for.
> 
> Understood regarding the implicit limits -- if a transaction has more calldata than `(blockGasLimit - 21000) / 4` zero bytes, its intrinsic gas exceeds the block gas limit, and it's rejected (`tx.Size()` encoded transaction size is close), correct?
> 
> Would it make sense to set `MaxSize` to something derived from the block gas limit (e.g. `s.curr.GasLimit / 4`) so the same transaction is rejected at the byte-size check before reaching the gas validation? AFAIK, failing fast is preferable, so does it make sense to catch something here, rather than deferring to the gas checks?

`intmath.CeilDiv(s.curr.GasLimit, params.TxDataNonZeroGasEIP2028)` ensures that it's just for efficiency and doesn't affect consensus in any way.

PR #256, opened 2026-03-03
test: fix fuzz caching

### LINKED ISSUE #209: `setup-go` not caching fuzz corpora in CI
See: https://github.com/ava-labs/strevm/blob/5faf8141974efdc39dca07b179f74c18db7a8718/.github/workflows/go.yml#L22-L23

The `go_fuzz` CI job was removed because `setup-go` wasn't properly caching fuzz corpora across runs.

Investigate whether newer versions of `actions/setup-go` handle this better, or set up a separate cache step for the fuzz corpus directory.

JonathanOppenheimer:
## Issue 

This PR reinstates the `go_fuzz` CI job that was removed in #46. The original job relied on setup-go's implicit caching of `GOCACHE` (which includes the fuzz subdirectory) but the fuzz corpus wasn't persisting across runs. This adds a dedicated `actions/cache` step with a per-function key independent of the `go.mod`, so the corpus accumulates reliably across CI runs.

Note that while there are environment variables for other Go caches, there's no go fuzz cache that would make this a lot easier -- see this longstanding issue: https://github.com/golang/go/issues/48901

This PR also fixes `require_fuzz_corpus` where the `${{ matrix.target }}` was consumed into the shell commands directly. I made both jobs use an `env: block`, which I believe is better practice and cleaner. 

## Testing 
 
I tested this works running the fuzz tests locally and verifying the cache was written to `$GOCACHE/fuzz/`. We see this also works in CI in the second workflow run (a reun) through `"Cache restored from key: fuzz-corpus-Linux-FuzzEffectiveGasTip-..."`. See the run [here](https://github.com/ava-labs/strevm/actions/runs/22635004856/job/65595900510?pr=256). 

Screenshot 2026-03-03 at 12 36 27 PM

We can also verify it works in CI in an entirely new run (a new commit, eabf536), where the primary key is missed, but the restore key is matched. 

Screenshot 2026-03-03 at 12 47 08 PM

----

Closes #209

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -89,29 +89,27 @@ jobs:
       - uses: actions/checkout@v4
       - id: find
         run: |
-          echo "targets=$(grep --recursive --include '**_test.go' -oP "^func \KFuzz[^(]+" | xargs jq -c -n '$ARGS.positional' --args)" >> $GITHUB_OUTPUT
+          targets=$(grep --recursive --include '**_test.go' -oP "^func \KFuzz[^(]+" | while IFS=: read -r file func; do
+            jq -c -n --arg p "$(dirname "$file")" --arg f "$func" '{package: $p, function: $f}'
+          done | jq -c -s '.')
+          echo "targets=${targets}" >> $GITHUB_OUTPUT
 
   go_fuzz:
     needs: [find_fuzz_targets]
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        target: ${{fromJSON(needs.find_fuzz_targets.outputs.targets)}}
+        include: ${{fromJSON(needs.find_fuzz_targets.outputs.targets)}}
     env:
+      PACKAGE: ${{ matrix.package }}
+      FUNCTION: ${{ matrix.function }}
+      CORPUS_DIR: ./${{ matrix.package }}/testdata/fuzz/${{ matrix.function }}
       # The actual corpus size is dependent on the test in question and how many
       # "interesting" cases can be found. This is an arbitrary lower bound,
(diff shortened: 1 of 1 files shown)


.github/workflows/go.yml
```diff
@@ -55,6 +53,33 @@ jobs:
...
+        uses: actions/setup-go@v5
+        with:
+          go-version-file: "go.mod"
+      - name: Restore fuzz corpus cache
+        uses: actions/cache@v4
+        with:
+          path: ~/.cache/go-build/fuzz
+          key: fuzz-corpus-${{ runner.os }}-${{ env.FUNCTION }}-${{ github.run_id }}
+          restore-keys: |
+            fuzz-corpus-${{ runner.os }}-${{ env.FUNCTION }}-
```
ARR4N:
```suggestion
            fuzz-corpus-${{env.PACKAGE}}-${{ env.FUNCTION }}
```

Why include the OS?

.github/workflows/go.yml
```diff
@@ -55,6 +53,33 @@ jobs:
           go-version-file: "go.mod"
       - run: go test ./... -race
 
+  go_fuzz:
```
ARR4N: Given the overlap between this and `require_fuzz_corpus`, please just add the `setup-go`, `cache`, and `go test -fuzz` steps of your `go_fuzz` to the end of the existing `require_fuzz_corpus`, and rename it `go_fuzz`. So it parses the target, ensures a baseline corpus, then fuzzes along with the cache.

I was considering just removing the required corpora but I like that everyone who downloads the repo has some baseline level of coverage when they just run vanilla `go test`.

PR #261, opened 2026-03-05
fix: remove duplicate logs

StephenButtolph:
There are two fixes to the logging done in `saexec/execution.go`:
1. When encountering a FATAL error, there would be two log entries. One logged by the `execute` function and one by the `processQueue` function. Now only the `processQueue` function reports errors.
2. Inside `execute`, we overwrote `logger` with `logger = logger.With(` for every tx iteration. This causes O(n^2) memory allocations.

Related (but non-functional) changes:
- `logger` was renamed to `log` to align with the rest of the codebase.
- `log := log.With(` was replaced with `blockLog := log.With(` in `sae/block_builder.go` to avoid any potential future O(n^2) memory allocations (and to align with the `txLog` and `opLog` that is performed in the rest of the function).

### DIFF
--- a/sae/rpc.go
+++ b/sae/rpc.go
@@ -255,7 +255,7 @@ func (s *netAPI) Version() string {
 
 type blockChainAPI struct {
 	*ethapi.BlockChainAPI
-	b *ethAPIBackend
+	b *apiBackend
 }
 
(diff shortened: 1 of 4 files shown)


JonathanOppenheimer (approved):
😄

PR #263, opened 2026-03-05
feat: stub avalanche rpcs

JonathanOppenheimer:
This PR adds stub RPC endpoints for Avalanche-custom APIs that need to be supported alongside the standard geth RPCs.                            

This list reflects the additional RPCs we decided as a team should be supported post-SAE. If anyone has additional RPCs they want to support, now is the time to discuss! 

eth namespace extensions (`sae/rpc_ext.go`):                                                                                               
  - `eth_getChainConfig`                                      
  - `eth_baseFee`
  - `eth_callDetailed`
  - `eth_suggestPriceOptions`
  - `eth_newAcceptedTransactions` (subscription)

For now all the methods are panic stubs, because I just want to get consensus on which endpoints we'll expose and how we do so.

Lastly, regarding the choice to keep the `eth_*` prefix for our custom RPCs, polluting the `eth_*` namespace. I am personally against this, and think we should break people's use of these RPCs, and `avax_*` prefix these to get rid of the pollution, but after talking to @StephenButtolph, he expressed that we can deprecate the modifications to the `eth_*` namespace and move the functions to a different namespace later if we so choose. But that for now it would be better to maintains backwards compatibility.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -120,11 +120,17 @@ jobs:
           go-version-file: "go.mod"
       - name: Get Go cache directory
         run: echo "GOCACHE=$(go env GOCACHE)" >> "$GITHUB_ENV"
+      - name: Compute cache keys
+        run: |
+          CORPUS_KEY="${{ matrix.package }}-${{ matrix.function }}_static-corpus_${{ hashFiles(format('{0}/testdata/fuzz/{1}/**', matrix.package, matrix.function)) }}"
+          COMMIT_KEY="${CORPUS_KEY}_commit_${{ github.sha }}"
+          echo "CORPUS_KEY=${CORPUS_KEY}" >> "$GITHUB_ENV"
+          echo "COMMIT_KEY=${COMMIT_KEY}" >> "$GITHUB_ENV"
       - name: Restore fuzz corpus cache
         uses: actions/cache@v4
         with:
           path: ${{ env.GOCACHE }}/fuzz
-          key: fuzz-corpus-${{ matrix.package }}-${{ matrix.function }}-${{ hashFiles(format('{0}/testdata/fuzz/{1}/**', matrix.package, matrix.function)) }}
+          key: ${{ env.COMMIT_KEY }}
           restore-keys: |
-            fuzz-corpus-${{ matrix.package }}-${{ matrix.function }}-
+            ${{ env.CORPUS_KEY }}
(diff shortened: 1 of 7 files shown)


sae/rpc_avax.go
```diff
@@ -0,0 +1,58 @@
+// Copyright (C) 2025-2026, Ava Labs, Inc. All rights reserved.
+// See the file LICENSE for licensing terms.
+
+package sae
+
+// Avalanche-specific "avax" namespace RPCs for atomic transactions. These are
+// served on a separate HTTP endpoint (/avax) using gorilla/rpc, matching the
+// pattern established in graft/coreth/plugin/evm/atomic/vm.
+//
+// The actual implementations will live in coreth but strevm provides the
```
JonathanOppenheimer: Right? I'm not sure exactly what we'll be keeping / deleting post graft / implementation.

sae/rpc_ext.go
```diff
@@ -0,0 +1,78 @@
...
+	"github.com/ava-labs/libevm/params"
+	"github.com/ava-labs/libevm/rpc"
+)
+
+// customAPI exposes Avalanche-custom extensions under the "eth" namespace:
+//   - eth_getChainConfig
+//   - eth_baseFee
+//   - eth_suggestPriceOptions
+//   - eth_callDetailed
+//   - eth_newAcceptedTransactions (subscription)
```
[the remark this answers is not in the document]
JonathanOppenheimer: I'll just delete this comment, but add it to your api support branch.

sae/rpc_ext.go
```diff
@@ -0,0 +1,71 @@
...
+	GasTip *hexutil.Big `json:"maxPriorityFeePerGas"`
+	GasFee *hexutil.Big `json:"maxFeePerGas"`
+}
+
+// PriceOptions groups slow/normal/fast gas-price suggestions.
+type PriceOptions struct {
+	Slow   *Price `json:"slow"`
+	Normal *Price `json:"normal"`
+	Fast   *Price `json:"fast"`
+}
```
StephenButtolph: I hate that these types pollute the `sae` namespace... Technically, we don't use them in any client implementation... So we could unexport them... I had originally thought it might be nice for coreth/subnet-evm to depend on these implementations... But it probably makes more sense to just unexport them here and live with the duplicate code until the coreth+subnet-evm rpc implementations are completely deleted.

StephenButtolph (approved):

PR #271, opened 2026-03-06
feat: add `GetChainConfig`,`BaseFee`, `SuggestPriceOptions` RPCs

JonathanOppenheimer:
## Summary

This PR implements `eth_getChainConfig`, `eth_baseFee`, and `eth_suggestPriceOptions` RPCs, which are required for compatibility with tooling like the Core wallet. 

- `eth_getChainConfig` does as it says haha. 
- `eth_baseFee` returns the worst-case upper bound on the next block's base fee, returning nil at genesis (matching coreth)
- `eth_suggestPriceOptions` returns three-tier gas pricing (slow/normal/fast), where each tier contains a tip and fee cap (`2*baseFee + tip`) with tip scaled at 95%/100%/105% respectively

I also added  `Estimator.EstimateNextBaseFee()` to the `gasprice` package, so the RPC doesn't need to reach into the `VM`, which also matches coreth. 

## Testing
- `TestGetChainConfig`: returns the expected chain config
- `TestBaseFee`: base fee matches worst-case bounds after a consensus round
- `TestSuggestPriceOptions`: validates three-tier pricing, fee cap formula, and tip ordering (slow ≤ normal ≤ fast)

### DIFF
--- a/sae/rpc_custom_test.go
+++ b/sae/rpc_custom_test.go
@@ -49,7 +49,7 @@ func TestSuggestPriceOptions(t *testing.T) {
 	b := sut.runConsensusLoop(t)
 
 	// This just asserts the round-tripping of the PriceOptions through the RPC.
-	// See rpc.TestNewPriceOptions for behavioral tests.
+	// See testing of [saerpc.NewPriceOptions] for behavioral tests.
 	tip, err := sut.rawVM.GethRPCBackends().SuggestGasTipCap(t.Context())
 	require.NoErrorf(t, err, "SuggestGasTipCap()")
 	doubleBaseFee := b.WorstCaseBounds().LatestEndTime.BaseFee().ToBig()


sae/rpc_ext.go
```diff
@@ -59,9 +66,54 @@ type priceOptions struct {
 	Fast   *price `json:"fast"`
 }
 
+// Tip-speed scaling constants, matching coreth's defaults.
+// See: github.com/ava-labs/avalanchego/blob/v1.14.1/graft/coreth/plugin/evm/config/config.go
+const (
+	slowTipPct = 95
+	fastTipPct = 105
+)
```
JonathanOppenheimer: I originally put these on the RPCConfig, but because a default (zero) value for these is problematic, I didn't like functions to return a set default if it was unset. If someone wants to change these, we can do so later IMO.

gasprice/estimator.go
```diff
@@ -317,6 +317,17 @@ func (e *Estimator) FeeHistory(
...
+// EstimateNextBaseFee returns the worst-case upper bound on the next block's
+// base fee. It returns nil when the last accepted block has no worst-case
+// bounds, which happens when it is the genesis block.
+func (e *Estimator) EstimateNextBaseFee() *big.Int {
+	bounds := e.backend.LastAcceptedBlock().WorstCaseBounds()
+	if bounds == nil {
+		return nil
+	}
+	return bounds.LatestEndTime.BaseFee().ToBig()
+}
```
StephenButtolph: I think this could just be a function defined in `rpc_ext.go`. I don't see how we gain anything by putting it here.

sae/rpc_ext.go
```diff
@@ -24,12 +26,17 @@ type customAPI struct {
...
-// BaseFee returns the worst-case base fee of the last accepted block.
+// BaseFee returns an upper-bound estimate of the base fee for the next block.
+// It returns nil if the estimate is unavailable, matching coreth's behaviour.
 func (c *customAPI) BaseFee(ctx context.Context) (*hexutil.Big, error) {
-	panic(errUnimplemented)
+	baseFee := c.b.EstimateNextBaseFee()
+	if baseFee == nil {
+		return nil, nil
+	}
+	return (*hexutil.Big)(baseFee), nil
```
StephenButtolph: This code is equivalent.
```suggestion
	return (*hexutil.Big)(c.b.EstimateNextBaseFee()), nil
```

sae/rpc_ext.go
```diff
@@ -59,9 +73,53 @@ type priceOptions struct {
...
 
+// Tip-speed scaling constants.
+const (
+	slowTipPct = 95
+	fastTipPct = 105
+)
+
+var (
+	minGasTip = big.NewInt(params.Wei)
+	big100    = big.NewInt(100)
```
StephenButtolph: pedantic, but we have been trying to define the vars as close to where they are used as possible. `big100` can be moved to on top of `scaleTip`.

sae/rpc_ext_test.go
```diff
@@ -0,0 +1,69 @@
...
+	bounds := b.WorstCaseBounds()
+	require.NotNil(t, bounds, "WorstCaseBounds()")
+
+	sut.testRPC(ctx, t, rpcTest{
+		method: "eth_baseFee",
+		want:   (*hexutil.Big)(bounds.LatestEndTime.BaseFee().ToBig()),
+	})
+}
+
+func TestSuggestPriceOptions(t *testing.T) {
```
StephenButtolph: Why can't we use `testRPC` for this?

StephenButtolph: Did my best in #272

sae/rpc_ext.go
```diff
@@ -52,16 +66,57 @@ type price struct {
...
+}
+
 // SuggestPriceOptions returns gas-price suggestions at three speed tiers.
 func (c *customAPI) SuggestPriceOptions(ctx context.Context) (*priceOptions, error) {
-	panic(errUnimplemented)
+	tip, err := c.b.SuggestGasTipCap(ctx)
+	if err != nil {
+		return nil, err
+	}
+	doubleBaseFee := c.estimateNextBaseFee()
```
alarso16: Why is this double the base fee?

sae/rpc/custom.go
```diff
@@ -24,12 +26,24 @@ type customAPI struct {
...
-	panic(errUnimplemented)
+	return c.b.ChainConfig()
 }
 
-// BaseFee returns the worst-case base fee of the last accepted block.
-func (c *customAPI) BaseFee(ctx context.Context) (*hexutil.Big, error) {
-	panic(errUnimplemented)
+// BaseFee returns an upper-bound estimate of the base fee for the next block.
+// It returns nil if the estimate is unavailable.
+func (c *customAPI) BaseFee(ctx context.Context) *hexutil.Big {
```
ceyonur: Why we changed the API signature?

sae/rpc_custom_test.go
```diff
@@ -0,0 +1,107 @@
...
+	})
+
+	b := sut.runConsensusLoop(t)
+	sut.testRPC(ctx, t, rpcTest{
+		method: "eth_baseFee",
+		want:   (*hexutil.Big)(b.WorstCaseBounds().LatestEndTime.BaseFee().ToBig()),
+	})
+}
+
+func TestNewPriceOptions(t *testing.T) {
```
StephenButtolph: Can we move this test inside of the `rpc` package?

PR #275, opened 2026-03-09
feat: `CallDetailed` and `NewAcceptedTransactions` RPCs  

JonathanOppenheimer:
This PR adds support for the last two custom RPCs: 
                                                                                                                     
- `eth_callDetailed`: performs the same call as `eth_call` but returns gas usage, error code, and revert data instead of just the return bytes.                                     
- `eth_newAcceptedTransactions`: WebSocket subscription that streams transactions as they are accepted by consensus (prior to execution). Supports both hash-only and full-tx modes via an optional `fullTx` parameter.

I wrote a test for each in `rpc_test_ext.go`: 
- `TestCallDetailed`: exercises successful calls, on-disk vs in-memory blocks, and reverting calls with decoded revert data
- `TestNewAcceptedTransactions`: verifies hash-only and full-tx subscription modes


sae/rpc_ext.go
```diff
@@ -56,8 +61,27 @@ type detailedExecutionResult struct {
 
 // CallDetailed performs the same call as eth_call, but returns gas usage and
 // error details instead of just the return data.
-func (c *customAPI) CallDetailed(ctx context.Context, args any, blockNrOrHash rpc.BlockNumberOrHash, overrides any) (*detailedExecutionResult, error) {
-	panic(errUnimplemented)
+func (c *customAPI) CallDetailed(ctx context.Context, args ethapi.TransactionArgs, blockNrOrHash rpc.BlockNumberOrHash, overrides *ethapi.StateOverride) (*detailedExecutionResult, error) {
```
JonathanOppenheimer: See https://github.com/ava-labs/avalanchego/blob/1a59a6f646ef18167dde10d24d4efbf05e3de177/graft/coreth/internal/ethapi/api_extra.go#L33

PR #280, opened 2026-03-12
refactor: abstract all RPC functionality out of `sae`

ARR4N:
The implementation of RPCs inside the `sae` package led to a tightly coupled monolith with no clear boundary between the API and the backing VM. This PR aims to change that by abstracting most of the code into a new `sae/rpc` package and block-specific functionality into the `blocks` package. Only a thin provider remains in the `sae` package, which provides only what is necessary for RPC backing, and in an explicit fashion.

`blocks/access.go` contains the new `Resolve*()` and `From*()` functions, which were originally `resolveBlockNumber()` and `readBy*()`. They provide explicit not-found errors instead of conforming to the idiosyncratic behaviour of geth's `rawdb` and RPC error (or lack thereof) handling.

All RPC backend implementations are now in `sae/rpc` in individual files that group methods by general intent. For the most part they remain entirely unchanged but the breaking up of the files makes it impossible for `git` to track. 

**The only implementation change** is `getReceipts()` in `sae/rpc/receipts.go`, which previously relied on the `sae.VM.settledBlockFromDB()` method. This was unnecessary and only done because we knew the block to be settled, but the new self-contained anonymous function simply retrieves the receipts.

Tests remain in `sae` to demonstrate that they are identical (save for one minor refactoring to use the abstracted identifiers) and that the refactoring did not break them. They SHOULD NOT be moved in the future as they are _integration_ tests of the VM<>RPC whole.

All other changes flowed out of necessity (e.g. moving `Executor.SignerForBlock()`) except for the renaming of `sae.VM.blocks` to `sae.VM.inConsensus` to better document what the `syncMap` holds. This was inspired by the new `blocks.InConsensus` interface, which it backs. This change provides a clearer definition of what `fromMem` function arguments were receiving.

### DIFF
--- a/blocks/access.go
+++ b/blocks/access.go
@@ -260,7 +260,7 @@ func FromNumberAndHash[T any](c Chain, hash common.Hash, rpcNum rpc.BlockNumber,
 	if rpcNum < 0 {
 		return nil, errors.New("named blocks not supported")
 	}
-	n := uint64(rpcNum)
+	n := uint64(rpcNum) //nolint:gosec // Non-negative check performed above
 	if b, ok := c.ConsensusCriticalBlock(hash); ok {
 		if b.NumberU64() != n {
 			return nil, fmt.Errorf("%w: found block number %d for hash %#x, expected %d", ErrNotFound, b.NumberU64(), hash, n)


blocks/access.go
```diff
@@ -0,0 +1,268 @@
...
+	return fromDB(c.DB(), hash, n)
+}
+
+// FromNumberAndHash behaves like [FromNumberOrHash] except that it accepts both
+// the number and hash as separate, required arguments. It verifies that any
+// [Block] found in consensus has the expected number.
+func FromNumberAndHash[T any](c Chain, hash common.Hash, rpcNum rpc.BlockNumber, fromConsensus Extractor[T], fromDB DBReaderWithErr[T]) (*T, error) {
+	if hash == (common.Hash{}) {
+		return nil, errors.New("empty block hash")
+	}
```
StephenButtolph: Relating back to the comment w.r.t. `GetBody`, we disallow the empty hash (from geth) but allow unresolved numbers (unlike geth). This isn't a change in behavior... But it is weird.

PR #281, opened 2026-03-12
refactor: prevent mutation of shared `uint256.Int` and `big.Int` values

JonathanOppenheimer:
This PR has to fixes to prevent @ARR4N's concern/FYI about returning pointers to big ints as they can be mutated by the caller, which corrupts them for everyone. 

1. `uint256.Int`: use value types instead of pointers. I also learned that the underlying representation of a uint256.Int is `[4]uint64`. 
2. `*big.Int`: clone at return boundaries to make mutation harmless. 

Note that for a nil check, `blocks.Block.BaseFee()` previously returned `nil` when the block hadn't been executed. It now returns the zero value `uint256.Int{}`. It doesn't appear that any caller relied on `nil` to distinguish `"not executed"` from "base fee is zero", as we have `Block.Executed()`. 

I did not change `txgossip/priority.go` as all the `*uint256.Ints` come from the external `txpool` API or `sae.uint256FromBig()` which returns `*uint256.Int` to set `txpool.LazyTransaction` fields -- if anyone has any thoughts on these I am open!

### DIFF
--- a/blocks/execution.go
+++ b/blocks/execution.go
@@ -211,7 +211,7 @@ func executionArtefact[T any](b *Block, desc string, get func(*executionResults)
 
 func (e *executionResults) executedByGasTime() *gastime.Time    { return e.byGas.Clone() }
 func (e *executionResults) executedByWallTime() time.Time       { return e.byWall }
-func (e *executionResults) cloneBaseFee() *uint256.Int          { return e.baseFee.Clone() }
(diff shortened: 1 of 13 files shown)


JonathanOppenheimer:
> Please do not commit before RPC abstraction.

I can change the base branch, and redo this work!

PR #289, opened 2026-03-23
fix: use worst case base fee

JonathanOppenheimer:
When the last accepted block has no worst-case bounds, `FeeHistory` and `estimateNextBaseFee` were falling back to the execution base fee (`BaseFee().ToBig()`) instead of the worst-case base fee (`EthBlock().BaseFee()`). 

The worst-case base fee is always >= the execution base fee, so without this, we could underestimate the next block's base fee.  I also added regression test that creates a block where the two fees diverge and asserts the worst-case value is returned.


ARR4N:
Deferring to @StephenButtolph on this.

PR #290, opened 2026-03-23
feat: add `CallDetailed()` RPC 

JonathanOppenheimer:
This PR adds support for `eth_callDetailed`, which performs the same call as `eth_call` but returns gas usage, error code, and revert data instead of just the return bytes.

I included a unit test, `TestCallDetailed` which exercises successful calls, on-disk vs in-memory blocks, and reverting calls with decoded revert data.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -55,7 +55,6 @@ linters:
     - testableexamples
     - testifylint
     - thelper
-    - tparallel
     - unconvert
     - unused
     - usestdlibvars
--- a/blocks/access.go
+++ b/blocks/access.go
@@ -12,7 +12,7 @@ import (
 	"github.com/ava-labs/libevm/ethdb"
 	"github.com/ava-labs/libevm/rpc"
 
-	"github.com/ava-labs/strevm/saedb"
+	"github.com/ava-labs/strevm/types"
 )
 
 type (
@@ -21,7 +21,7 @@ type (
 		ConsensusCritical
 		Frontier
 		DB() ethdb.Database
-		XDB() saedb.ExecutionResults
+		XDB() types.ExecutionResults
 	}
 
 	// ConsensusCritical blocks are currently in use by a consensus mechanism,
--- a/blocks/block.go
+++ b/blocks/block.go
@@ -22,7 +22,6 @@ import (
 	"go.uber.org/zap"
 
 	"github.com/ava-labs/strevm/proxytime"
-	"github.com/ava-labs/strevm/saedb"
 	saetypes "github.com/ava-labs/strevm/types"
 )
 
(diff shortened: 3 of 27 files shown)


sae/rpc/custom.go
```diff
@@ -48,6 +50,10 @@ func (c *customAPI) estimateNextBaseFee() *big.Int {
 	return bounds.LatestEndTime.BaseFee().ToBig()
 }
 
+// RevertErrCode is the JSON-RPC error code for execution reverts, matching the
+// value returned by [ethapi.RevertError.ErrorCode].
+const RevertErrCode = 3
```
StephenButtolph: Is there a reason we define this constant here? It seems that in prior iterations it was used in `CallDetailed` (so it was added). Then it was used in the tests (so it was exported). But now we don't use it in `CallDetailed` at all.

Can we move this definition to the test?

sae/rpc/custom.go
```diff
@@ -58,8 +64,28 @@ type DetailedExecutionResult struct {
...
+		UsedGas:    result.UsedGas,
+		ReturnData: result.ReturnData,
+	}
+	// Revert data is checked first because [ethapi.NewRevertError] ABI-decodes
+	// the reason, which we provide.
+	if errors.Is(result.Err, vm.ErrExecutionReverted) {
+		e := ethapi.NewRevertError(result.Revert())
+		reply.Err = e.Error()
+		reply.ErrCode = e.ErrorCode()
+	} else if result.Err != nil {
```
StephenButtolph: No action required for this PR, but it is worth noting that changing to inspect the error rather than checking if `len(result.Revert()) > 0` is a bug fix that currently impacts coreth + subnet-evm. Not sure if we want to fix this there as well.

sae/rpc/custom.go
JonathanOppenheimer: Yes - it also then does not make any sense to export it anymore as the prior justification doesn't make sense IMO.

sae/rpc/custom.go
JonathanOppenheimer: I think it makes sense to. I can create an issue to do so. https://github.com/ava-labs/avalanchego/issues/5139

sae/rpc_custom_test.go
```diff
@@ -98,3 +106,126 @@ func TestNewAcceptedTransactions(t *testing.T) {
...
+	const revertWith = 12345
+	revertAsPanic := slices.Concat(
+		crypto.Keccak256([]byte("Panic(uint256)"))[:4],
+		uint256.NewInt(revertWith).PaddedBytes(32),
+	)
+
+	noBalance := common.Address{'b', 'a', 'n', 'k', 'r', 'u', 'p', 't'}
+	latest := rpc.LatestBlockNumber.String()
+
+	sut.testRPC(ctx, t, []rpcTest{
```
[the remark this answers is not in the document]
JonathanOppenheimer: I have just added this exactly.

PR #278, opened 2026-03-11
fix: make execution artefact accessors blocking    

JonathanOppenheimer:
This PR make functions that access execution-artefacts blocking, so they wait on the block's executed channel instead of returning zero values. A warning is logged if blocking exceeds `MaxQueueWallTime`. 

The end goal is that we can remove the need for the explicit `Executed()` guard and `ErrNotExecuted` in `StateAndHeaderByNumberOrHash`, as we can just rely on the blocking behavior directly.

PR #298, opened 2026-03-25
refactor: reject non-canonical blocks in RPC queries

JonathanOppenheimer:
I broke this out from https://github.com/ava-labs/strevm/pull/278 for ease of review and merge. 

This PR makes it so that RPC readers now enforce `RequireCanonical`, which rejects verified-but-not-accepted blocks with `ErrNonCanonicalBlock`. This prevents state queries from blocking indefinitely on execution art**E**facts that may never materiali**S**e.

### DIFF
--- a/sae/rpc_stateful_test.go
+++ b/sae/rpc_stateful_test.go
@@ -36,12 +36,11 @@ func TestStateQueryOnNonCanonicalBlock(t *testing.T) {
 	ctx, sut := newSUT(t, 1)
 	b := unwrap(t, sut.createAndVerifyBlock(t, sut.lastAcceptedBlock(t)))
 
-	wantErr := testerr.Contains(blocks.ErrNonCanonicalBlock.Error())
 	sut.testRPC(ctx, t, []rpcTest{
 		{
 			method:  "eth_getBalance",
 			args:    []any{sut.wallet.Addresses()[0], rpc.BlockNumberOrHashWithHash(b.Hash(), false)},
-			wantErr: wantErr,
+			wantErr: testerr.Contains(blocks.ErrNonCanonicalBlock.Error()),
 		},
 		{
 			method: "eth_getBlockByHash",


StephenButtolph (commented):
```golang
func (b *backend) getReceipts(numOrHash rpc.BlockNumberOrHash) (types.Receipts, *types.Block, error) {
	// If the block isn't available from the [blocks.InConsensus] of the
	// [blocks.Chain] then it can only have receipts if it's settled and
	// therefore must be canonical.
	numOrHash.RequireCanonical = true
```

Do we need this setting anymore?

sae/rpc/readers.go
```diff
@@ -26,15 +26,20 @@ func notFoundIsNil[T any](x *T, err error) (*T, error) {
...
 	return notFoundIsNil(blocks.FromNumber(c, n, read.WithNilErr()))
 }
 
 func readByHash[T any](c Chain, hash common.Hash, fromMem blocks.Extractor[T], fromDB blocks.DBReader[T]) (*T, error) {
-	return notFoundIsNil(blocks.FromHash(c, hash, fromMem, fromDB.WithNilErr()))
+	return notFoundIsNil(blocks.FromHash(c, hash, true, fromMem, fromDB.WithNilErr()))
 }
 
 func readByNumberOrHash[T any](c Chain, blockNrOrHash rpc.BlockNumberOrHash, fromMem blocks.Extractor[T], fromDB blocks.DBReaderWithErr[T]) (*T, error) {
+	blockNrOrHash.RequireCanonical = true
```
StephenButtolph: Why do we set `RequireCanonical = true` everywhere? Can't we just ignore the field entirely?

JonathanOppenheimer:
> ```go
> func (b *backend) getReceipts(numOrHash rpc.BlockNumberOrHash) (types.Receipts, *types.Block, error) {
> 	// If the block isn't available from the [blocks.InConsensus] of the
> 	// [blocks.Chain] then it can only have receipts if it's settled and
> 	// therefore must be canonical.
> 	numOrHash.RequireCanonical = true
> ```
> 
> Do we need this setting anymore?

ah no this is leftover we do not need it anymore.

PR #303, opened 2026-03-30
refactor: unexport `saedb` helpers

alarso16:
There's a few functions that are exported, but were only used in testing. They can be unexported

### DIFF
--- a/sae/recovery_test.go
+++ b/sae/recovery_test.go
@@ -196,20 +196,16 @@ func TestRecoverSimple(t *testing.T) {
 			// where the settled state was written to disk.
 			t.Run("unavailable_outside_window", func(t *testing.T) {
 				lastSettled := sut.rawVM.last.settled.Load().NumberU64()
-				committedHeight := saedb.LastCommittedTrieDBHeight(lastSettled)
+				// height at which the settled state was committed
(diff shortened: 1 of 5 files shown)


sae/recovery_test.go
```diff
@@ -196,20 +196,16 @@ func TestRecoverSimple(t *testing.T) {
 			// where the settled state was written to disk.
 			t.Run("unavailable_outside_window", func(t *testing.T) {
 				lastSettled := sut.rawVM.last.settled.Load().NumberU64()
-				committedHeight := saedb.LastCommittedTrieDBHeight(lastSettled)
+				// height at which the settled state was committed
+				committedHeight := saedb.CommitTrieDBEvery * (lastSettled / saedb.CommitTrieDBEvery) // round down to [saedb.CommitTrieDBEvery] interval
```
alarso16: I think this might be missing the spirit of the problem, since rather than using the helper method, we reimplement it. However, I do think we need to keep the function of this subtest (ensuring there's no memory leak on startup). For Firewood specifically, there's no way to know which state root is on disk.

My only other idea is to open a TrieDB before restarting the VM and lookback through blocks until a state is found. This block is used here.

alarso16:
I think this will be necessary for State Sync - I'll re-open if necessary

PR #304, opened 2026-03-30
refactor(gastime): remove hooks import

alarso16:
Removing the hooks imports in gastime makes its implementation more intuitive, and not depend on block headers. The naming may now be somewhat overkill. The motivation is to be able to provide the gastime into `BuildBlock` to add the minimum necessary data for state sync through this hook simply, rather than providing a custom type that takes in:

- Excess
- Seconds
- Fraction

### DIFF
--- a/blocks/execution_test.go
+++ b/blocks/execution_test.go
@@ -25,7 +25,6 @@ import (
 
 	"github.com/ava-labs/strevm/cmputils"
 	"github.com/ava-labs/strevm/gastime"
-	"github.com/ava-labs/strevm/hook"
 	"github.com/ava-labs/strevm/saetest"
 	saetypes "github.com/ava-labs/strevm/types"
 )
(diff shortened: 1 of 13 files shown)


gastime/gastime.go
```diff
@@ -62,15 +61,14 @@ func New(at time.Time, target, startingExcess gas.Gas, gasPriceConfig hook.GasPr
 	}, nil
 }
 
-// SubSecond scales the value returned by [hook.Points.SubSecondBlockTime] to
-// reflect the given gas rate.
-func SubSecond(hooks hook.Points, hdr *types.Header, rate gas.Gas) gas.Gas {
+// SubSecond scales [time.Duration] to reflect the given gas rate.
+func SubSecond(subSec time.Duration, rate gas.Gas) gas.Gas { //nolint:staticcheck // subSec intentionally communicates that the value is < time.Second
 	// [hook.Points.SubSecondBlockTime] is required to return values in
```
JonathanOppenheimer: This is no longer true after your refactor -- probably should update this whole comment.

blocks/execution_test.go
```diff
@@ -25,7 +25,6 @@ import (
 
 	"github.com/ava-labs/strevm/cmputils"
 	"github.com/ava-labs/strevm/gastime"
-	"github.com/ava-labs/strevm/hook"
 	"github.com/ava-labs/strevm/saetest"
 	saetypes "github.com/ava-labs/strevm/types"
```
JonathanOppenheimer: Aren't we supposed to alias the import that is external? so libevmtypes? or am I getting it backwards

JonathanOppenheimer (approved):
Seems straightforward enough to me.

types/types.go
```diff
@@ -29,3 +32,37 @@ type (
...
+	MinPrice gas.Price
+	// StaticPricing is a flag indicating whether the gas price should be static
+	// at the minimum price.
+	StaticPricing bool
+}
+
+var (
+	errTargetToExcessScalingZero = errors.New("targetToExcessScaling must be non-zero")
+	errMinPriceZero              = errors.New("minPrice must be non-zero")
+)
```
powerslider, line 54: I would move the var block at the top of the file.

PR #315, opened 2026-04-06
test(sae): only sync mempool once

### LINKED ISSUE #306: Fix `TestRecoverFromDatabase` flake
See https://github.com/ava-labs/avalanchego/actions/runs/23820326319/job/69430698753?pr=5155

alarso16:
Resolves #306. The test logs show:
```
--- FAIL: TestRecoverFromDatabase (57.17s)
    recovery_test.go:72: 
        	Error Trace:	/Users/runner/work/avalanchego/avalanchego/vms/saevm/sae/vm_test.go:376
        	            				/Users/runner/work/avalanchego/avalanchego/vms/saevm/sae/vm_test.go:360
        	            				/Users/runner/work/avalanchego/avalanchego/vms/saevm/sae/vm_test.go:398
        	            				/Users/runner/work/avalanchego/avalanchego/vms/saevm/sae/vm_test.go:419
        	            				/Users/runner/work/avalanchego/avalanchego/vms/saevm/sae/vm_test.go:433
        	            				/Users/runner/work/avalanchego/avalanchego/vms/saevm/sae/vm_test.go:443
        	            				/Users/runner/work/avalanchego/avalanchego/vms/saevm/sae/recovery_test.go:72
        	Error:      	Condition never satisfied
        	Test:       	TestRecoverFromDatabase
        	Messages:   	all of txs [[]] to in mempool
FAIL
```

Notice that the tx set is empty. This means that we aren't actually finishing the callback, since there's nothing to assert. I believe this is caused by concurrent calls to `txpool.TxPool.Sync`, since looking at the worker loop, there's a logical race in boolean updating with concurrent calls. I attempted to verify this by editing `syncMempool` to look like:
```go
func (s *SUT) syncMempool(tb testing.TB) {
	tb.Helper()
	var _ txpool.TxPool // maintain import for [comment] rendering
	p := s.rawVM.mempool.Pool
	var wg sync.WaitGroup
	for range 2 {
		wg.Go(func() {
			require.NoErrorf(tb, p.Sync(), "%T.Sync()", p)
		})
	}
	wg.Wait()
}
```
We get the same error almost deterministically.

The correct fix is to only call `syncMempool` once. It's somewhat difficult to verify that this will certainly fix the flake seen when grafting, but I believe it will. Either way, this is still safe, since even though the txs might not be immediately added to the pool with `eth_sendRawTransaction`, they are available to be added as soon as a reset occurs (which is forced by `txpool.TxPool.Sync`).

### DIFF
--- a/sae/vm_test.go
+++ b/sae/vm_test.go
@@ -95,6 +95,8 @@ type SUT struct {
 	hooks   *hookstest.Stub
 	logger  *saetest.TBLogger
 
+	mempoolLock sync.Mutex // must be held when calling [txpool.TxPool.Sync]
+
 	validators *validatorstest.State
 	sender     *enginetest.Sender
 }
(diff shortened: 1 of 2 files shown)


sae/vm_test.go
```diff
@@ -363,21 +365,28 @@ func (s *SUT) addToMempool(tb testing.TB, txs ...*types.Transaction) {
...
+	defer s.mempoolLock.Unlock()
 	require.NoErrorf(tb, p.Sync(), "%T.Sync()", p)
 }
 
-// requireInMempool requires that the transaction with the specified hash is
-// eventually in the mempool. It calls [SUT.syncMempool] before every check.
+// requireInMempool requires that each transaction with the specified hashes are
+// eventually in the mempool. Internally calls [SUT.syncMempool], which holds
+// [SUT.mempoolLock].
 func (s *SUT) requireInMempool(tb testing.TB, txs ...common.Hash) {
```
StephenButtolph: I don't think we need to change this function at all now (both the comment and the implementation).

PR #318, opened 2026-04-07
test: Remove all uses of `txpool.TxPool.Sync()`

alarso16:
`txpool.TxPool.Sync` is inherently broken, since if there are any pending transactions currently in the mempool, the internal `noncer` isn't accurately recreated. Although we could make a large diff in the legacypool to fix this test-only code, it would be simpler to just wait until the tx is added via a subscription.

PR #320, opened 2026-04-08
fix: resolve `TestImmediateReceipts` flake

powerslider:
#### Code flow 
- The test builds a block with two txs, a simple transfer (`tx[0]`) and a call to a blocking precompile (`tx[1]`).
- After `Accept()`, the block is enqueued for async execution in a separate goroutine.
- The executor processes txs sequentially. `tx[0]` completes and its receipt is published immediately, then `tx[1]` enters the blocking precompile and waits on a channel.
- The test queries `tx[0]`'s receipt via RPC (which blocks until the receipt is ready), then asserts `b.Executed() == false`, expecting the precompile to still be blocking `tx[1]`.

#### The race condition
- When the RPC call returns, we know `tx[0]`'s receipt is ready, but there's no guarantee the execution goroutine has actually entered the precompile yet. 
- Under CI load, the goroutine could finish both txs before the assertion runs.

#### Proposed fix
- `withBlockingPrecompile` now returns a channel that closes when the precompile is entered.
- The test waits on it before asserting, so `b.Executed() == false` is deterministic.
- Add an `entered` channel to `withBlockingPrecompile` that closes when the precompile starts blocking.
- `TestImmediateReceipts` now waits on this channel before asserting, guaranteeing execution cannot complete.
- Test with `-count=1000` with no failures.

### DIFF
--- a/sae/rpc_receipts_test.go
+++ b/sae/rpc_receipts_test.go
@@ -11,11 +11,12 @@ import (
 	"github.com/ava-labs/libevm/core/types"
 	"github.com/ava-labs/libevm/params"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestImmediateReceipts(t *testing.T) {
 	blocking := common.Address{'b', 'l', 'o', 'c', 'k'}
-	opt, unblock := withBlockingPrecompile(blocking)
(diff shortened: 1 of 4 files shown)


StephenButtolph:
#318 is addressing the root cause here.

